PR #6, opened 2021-04-12
Fix for animations not playing more than 2 frames

srobertson421:
Fixed issue where animations with more than 2 frames wouldn't play beyond 2 frames. Only had to swap one line to allow for a variable length on the anim array.

### DIFF
--- a/src/kaboom.js
+++ b/src/kaboom.js
@@ -2907,7 +2907,7 @@ function sprite(id, conf = {}) {
 			if (this._animTimer >= this.animSpeed) {
 				// TODO: anim dir
 				this.frame++;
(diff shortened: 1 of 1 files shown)


slmjkdbtl:
Hi! Sorry this is not communicated well in the docs, the anim declaration actually specifies the frame range instead of all the frame, so

```
anims: {
    run: [0, 3],
    jump: [4, 4],
},
```

`run` will play from frame `0` to frame `3`, so there'll only be 2 items in these particular arrays

srobertson421:
Ah gotcha! That makes more sense, thanks for the quick response! I'll close this then

PR #11, opened 2021-04-14
Removed extra argument for `rgb` in documentation

JohnathonNow:
Just noticed that `rgb` had the `a` argument in the docs.


slmjkdbtl:
nice catch!

PR #13, opened 2021-04-14
Add kaboom global to site guide

rish:
Due to `kaboom.global()` being removed from the `genGame.js` script in this commit:

https://github.com/replit/kaboom/commit/e480fb5bf5c252b36e7325e79f83dc9c2f8d8c38#diff-0ddd15db301bfe15cb6bda312966ae5695dae3bd45a7fdddef4d50246e230837L63

The site guide doesn't load the sprites etc.

I've replicated what you have done in another commit for the examples, but it needs to be done for the interactive guide too.

Let me know if you need any other changes.

Many thanks!

### DIFF
--- a/site/pub/guide/10.js
+++ b/site/pub/guide/10.js
@@ -1,5 +1,7 @@
 // TALK: let's add a background image
 
+kaboom.global();
+
 loadSprite("bg", "/pub/img/bg.png");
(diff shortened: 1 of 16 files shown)


slmjkdbtl:
oh right, nice catch, thank you!

That should be enough for the fix, i'll merge.

PR #18, opened 2021-04-15
Include key when calling keypress callback functions.

charlestudor:
Hi! Brilliant library, it's lots of fun to make stuff with!

I noticed if you add the pressed key as an argument to the callback functions it allows for some powerful functionality like the ability to make quick input/dialog prompts to the user such as "Enter your name...".

Example code:
```
keyPress([ASCII_CHARS], (key) => {
    if (gameState.promptOpen) {
      promptInput(key);
    }
  });
```

Perhaps there is already an easy way to do this though that I just missed!

Many Thanks!

### DIFF
--- a/src/kaboom.js
+++ b/src/kaboom.js
@@ -2545,25 +2545,25 @@ function start(name, ...args) {
 			// run input checks & callbacks
 			for (const e of scene.events.keyDown) {
 				if (keyIsDown(e.key)) {
-					e.cb();
+					e.cb(e.key);
 				}
 			}
 
 			for (const e of scene.events.keyPress) {
 				if (keyIsPressed(e.key)) {
-					e.cb();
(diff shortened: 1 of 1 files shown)


slmjkdbtl:
There's actually a function called `charInput` which is used in this example here https://kaboomjs.com/examples#text, but just realized it's not documented yet, sorry. But yeah it does make more sense to pass the keys.

charlestudor:
Ah so there is! No worries, keep up the great work! 👍

slmjkdbtl:
FYI: Just realized I forgot about this and actually removed this in `v0.5.0` for some reason. I'll add it back later!

slmjkdbtl:
but your example looks like character input, there's a dedicated
```
charInput((ch) => {
	input.text += ch;
});
```
now

PR #22, opened 2021-04-16
Fix drive example car animation reference

rish:
The car animation is broken on the example due Aseprite support being ported over to `loadAseprite()` in kaboom version `0.2.0` instead of using it inside of a parameter inside of `loadSprite()`.

Animation now runs succesfully.

Also updated the documentation for `loadAseprite()`as well.


slmjkdbtl:
thanks!

PR #29, opened 2021-04-21
restore sprites required by interactive guide

dy-dx:
The guide is currently broken: https://kaboomjs.com/guide#5

Looks like these sprites were moved in https://github.com/replit/kaboom/commit/eefbe0beaf331c7f1642043b98c98c4293f92bcc


slmjkdbtl:
Thanks for the report! Actually I'd prefer to keep the example images under the `examples/`, and specify the path in the guide code, it's fixed in a9b8e10 (hefty) along with some other changes that require updating all the guide code files

PR #30, opened 2021-04-24
Updated README and added module version of kaboom.js

EpicGamer007:
At the moment, you declare `kaboom` as part of the "global" scope setting `window.kaboom = [...]`. I kept that version of the file in my fork but also added a `.mjs` version for it. All I changed was line 78 where instead of `window.kaboom = (args = {}) => {[...]}`, I added `export default (args = {}) => {[...]}`. I also changed the README to say you can `import` kaboom rther than including another script tag.


README.md
```diff
@@ -9,6 +9,11 @@ kaboom.js is a JavaScript library that helps you make games fast and fun!
 ```html
 <script src="https://kaboomjs.com/lib/0.4.0/kaboom.js"></script>
 <script type="module">
+
```
slmjkdbtl, line 12: Nice, but I'd put the the usage text under README#Usage section, this example is supposed to be the minimal to copy paste and does not have to show all usages.

slmjkdbtl:
Kaboom will soon be using a bundler to generate dist files, but happy to manually support now. Left a comment above about moving the usage text to `README#Usage`, another thing is I'd not put one under `src/` (to make the source dir cleaner, that means no `.mjs` on `/dev` version, but soon after moving to generated dist files, we'll use a `dist/` for dev distributing and support `mjs`)

slmjkdbtl:
Look good, thanks!

PR #39, opened 2021-04-29
Allow overlap and collide to work with multiple layers

srobertson421:
added logic to allow an array of layer strings to be passed to overlap and collide to allow multiple layers to overlap and collide.

Changed a few function signatures so am very open to figuring out a cleaner method so that it's not as breaking of a change.

### DIFF
--- a/src/kaboom.js
+++ b/src/kaboom.js
@@ -2305,18 +2305,18 @@ function render(tag, cb) {
 }
 
 // add an event that runs with objs with t1 collides with objs with t2
-function collides(t1, t2, f) {
+function collides(t1, t2, al, f) {
 	action(t1, (o1) => {
-		o1._checkCollisions(t2, (o2) => {
+		o1._checkCollisions(t2, al, (o2) => {
 			f(o1, o2);
 		});
 	});
 }
 
 // add an event that runs with objs with t1 overlaps with objs with t2
-function overlaps(t1, t2, f) {
+function overlaps(t1, t2, al, f) {
 	action(t1, (o1) => {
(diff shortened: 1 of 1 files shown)


srobertson421:
example of new signature would be:

```javascript
// Global overlap
overlap('tag1', 'tag2', ['bg', 'game'], (tag1, tag2) => {
  // overlap code
});

// Entity overlap
const player = add([
  sprite('player', '/path/to/sprite.png'),
  layer('game')
]);

player.overlap('spike', ['bg', 'game'], (player, spikeTile) => {
  // overlap code
});
```

slmjkdbtl:
Hmm that'd be a breaking change, I'm thinking about just not checking for layers all together, right now there's no escape hatch except make stuff the same layer, people can always check layers themselves if they wish

srobertson421:
Ya definitely don't want to make it harder than it needs to be. How would one go about checking layers manually?

slmjkdbtl:
Right now if you give it a `layer()` comp it'll assign a `.layer` field the obj.

another way to get around this for now is
```js
action("t1", (o1) => {
    every("t2", (o2) => {
        if (o1.isCollided(o2)) {
            // do stuff...
        }
    });
});
```

the reason behind different layer won't collide is I imagined layers to be conceptionally different worlds like UI, game scene, and backgrounds that won't really to interact with each other, so we don't pay the expense of checking the collision between a health bar and a bullet, but there might be better ways to define layers

triptych:
Should this be closed as you're not going to go with this strategy?

PR #48, opened 2021-05-23
center the game over text

luizbills:


slmjkdbtl:
Actually you'll have to update all subsequent files, but nvm it's handled in f01d4db, thanks for the notice!

PR #84, opened 2021-06-13
Comment Added

Aayam555:
I will be writing comments in these examples in the future.
For now I wrote comment for 1 file

So, Please put this in the project

### DIFF
--- a/examples/audio.js
+++ b/examples/audio.js
@@ -1,3 +1,5 @@
+
+// Initializing kaboom
 kaboom({
 	global: true,
 	fullscreen: true,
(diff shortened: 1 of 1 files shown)


triptych:
Not a fan of so much inline comments. Seems like it would be better to include a readme, or post a walkthrough with the comments instead of including so much in the source.

triptych:
I recommend closing this as "wont fix" .

slmjkdbtl:
Yeah sorry I think these comments are a bit messy, I'm planning to add better comments to all examples myself, I'll keep open for now as a reminder until that's finished

slmjkdbtl:
Added comments to audio example: 44737bf

Issue #80, opened 2021-06-13
Bug in flappy bird demo - No ceiling


coreysnodgrass:
I was able to click the spacebar a bunch and fly over all the pipes and beat the game. It seems like there should be an upper level preventing you from doing that.

I used chrome v 91

coreysnodgrass:
#85 - I created a PR for this.

slmjkdbtl:
Hey! Thanks for the PR, but actually I'd prefer to have it just die if it's too high up: 8c9ee81 (also the `0.001` height rect might not trigger collision because the birdy might move through it between frames)

PR #85, opened 2021-06-15
Added a ceiling for flappy bird so that you can't fly above everything

coreysnodgrass:
Well, I added a thin rectangle that is solid so that flappy bird wouldn't go above it. Was this in the original game? idk, but it seems odd to just fly over everything and get points for it.


slmjkdbtl:
replied in #80

PR #90, opened 2021-07-04
FIX typeError cannot redefine property width

RilDev:
There was an error in the Button example:

kaboom-button-example-typeerror-width

So I fixed it, and it now works as intended:

kaboom-button-example-typeerror-width-fixed


slmjkdbtl:
thx

PR #91, opened 2021-07-04
examples/rpg: fix error resolve is not a function

RilDev:
Fix error player.resolve() is not a function:

kaboom-player-example-resolve-is-not-a-function

After the fix, the example works as intended:

kaboom-rpg-example-fix-bug-resolve-not-a-function

### DIFF
--- a/examples/rpg.js
+++ b/examples/rpg.js
@@ -144,7 +144,7 @@ scene("main", (levelIdx) => {
 	}
 
 	player.action(() => {
-		player.resolve();
+		player.pushOutAll();
 	});
 
 });


slmjkdbtl:
`resolve()` get's renamed to `pushOutAll()` in master because it's a vague name and planning to get that less called and pack collision resolution into `.move()`, but wait it works? you're supposed to be able to move out of walls if you remove that 🤔

triptych:
Perhaps related to the odd roll back of code we saw with the tiling backgrounds example?

RilDev:
Hi, sorry for the late reply!

@slmjkdbtl you are right, the player was able to go through walls, I renamed the function to `pushOutAll()` and it worked fine!

PR #114, opened 2021-07-28
Make component functions accept parent  component as first argument

vgautamm:
Motivation: Currently when making custom components, we rely on "this" to access the properties of parent component. 

Since "this" can be confusing to beginners (including myself), having component as a first argument will always will make things clearer, it will also make the functions easier to test.

Here is a sample example

Before=
```
function big() {
	let timer = 0;
	let isBig = false;
	return {
		isBig() {
			return isBig;
		},
		smallify() {
			this.scale = vec2(1);
			timer = 0;
			isBig = false;
		},
		biggify(time) {
			this.scale = vec2(2);
			timer = time;
			isBig = true;
		},
	};
}

```

Proposed change
```
function big() {
	let timer = 0;
	let isBig = false;
	return {
		isBig() {
			return isBig;
		},
		smallify(comp) {
			comp.scale = vec2(1);
			timer = 0;
			isBig = false;
		},
		biggify(comp, time) {
			comp.scale = vec2(2);
			timer = time;
			isBig = true;
		},
	};
}

```

### DIFF
--- a/src/kaboom.ts
+++ b/src/kaboom.ts
@@ -470,7 +470,7 @@ function add(comps: Comp[]): GameObj {
 
 				// event / custom method
 				if (typeof comp[k] === "function") {
-					const func = comp[k].bind(this);
+					const func = comp[k].bind(this, obj);
 					if (COMP_EVENTS.has(k)) {
 						this.on(k, func);
 						continue;


vgautamm:
Since `this.bind(this, obj)` returns a partially applied function with the component applied as a first argument, it solves the problem.

slmjkdbtl:
This will be a big breaking change that requires more consideration. 

> it will also make the functions easier to test

I wonder what are you referring to by easier to test?

slmjkdbtl:
One thing is

```js
function myComp() {
    return {
        myMethod(obj, arg1, arg2) { // looks like accepts 3 arguments
            // ...
        }
    };
}

const obj = add([
    myComp(),
]);

obj.myMethod(arg1, arg2); // actually accepts 2 arguments
```

on the surface there's an imbalance on argument numbers in function def and function call which might cause some confusion.

I admit it's a bit confusing kaboom components treats `this` in a kinda special way, the mental model is

> `this` in component functions always refers to its host game object

to me it's pretty clear, wonder what other people think about these 2 approaches.

PR #118, opened 2021-08-01
Add complex type for add method

Wozza365:
Creates a dynamic type based on what's passed into the kaboom.add method.

This will merge the type of each object in an array and return that combined object.

It's often difficult to prove TypeScript changes, but here's an example in screenshots:

Before:
image
Limited list of suggestions - Only GameObj properties are suggested.

image
No valid IntelliSense for functions on any of the Comp's passed into the add method.

After: 
image
Full list of IntelliSense suggestions

image
Correct IntelliSense for available methods.

This is my first time using kaboom, so bear with me. Open to feedback :)

### DIFF
--- a/src/kaboom.ts
+++ b/src/kaboom.ts
@@ -422,15 +422,22 @@ const COMP_EVENTS = new Set([
 	"inspect",
 ]);
 
+type UnionToIntersection<U> =
+  (U extends any ? (k: U) => void : never) extends ((k: infer I) => void) ? I : never
+type Defined<T> = T extends any ? Pick<T, { [K in keyof T]-?: T[K] extends undefined ? never : K }[keyof T]> : never;
+type Expand<T> = T extends infer U ? { [K in keyof U]: U[K] } : never;
+
+type AddParameters = (PosComp | ScaleComp | RotateComp | ColorComp | OriginComp | LayerComp | AreaComp | SpriteComp | BodyComp | TextComp);
+
 // TODO: make tags also comp?
-function add(comps: Comp[]): GameObj {
+function add<T extends AddParameters>(comps: ReadonlyArray<T>): Expand<UnionToIntersection<Defined<T>>> & GameObj {
 
 	const compStates = {};
 	const customState = {};
 	const events = {};
 	const tags = [];
 
-	const obj: GameObj = {
+	const obj = {
 
 		_id: null,
 		hidden: false,
@@ -609,7 +616,7 @@ function add(comps: Comp[]): GameObj {
 		}
 	}
 
-	return obj;
+	return obj as unknown as Expand<UnionToIntersection<Defined<T>>> & GameObj;
 
 }
 


src/kaboom.ts
```diff
@@ -422,15 +422,22 @@ const COMP_EVENTS = new Set([
 	"inspect",
 ]);
 
+type UnionToIntersection<U> =
+  (U extends any ? (k: U) => void : never) extends ((k: infer I) => void) ? I : never
+type Defined<T> = T extends any ? Pick<T, { [K in keyof T]-?: T[K] extends undefined ? never : K }[keyof T]> : never;
+type Expand<T> = T extends infer U ? { [K in keyof U]: U[K] } : never;
+
+type AddParameters = (PosComp | ScaleComp | RotateComp | ColorComp | OriginComp | LayerComp | AreaComp | SpriteComp | BodyComp | TextComp);
```
Wozza365, line 430: There might be some missing Comp objects here, they should be added to this list.

src/kaboom.ts
```diff
@@ -422,15 +422,22 @@ const COMP_EVENTS = new Set([
 	"inspect",
 ]);
 
+type UnionToIntersection<U> =
+  (U extends any ? (k: U) => void : never) extends ((k: infer I) => void) ? I : never
+type Defined<T> = T extends any ? Pick<T, { [K in keyof T]-?: T[K] extends undefined ? never : K }[keyof T]> : never;
+type Expand<T> = T extends infer U ? { [K in keyof U]: U[K] } : never;
```
Wozza365, line 428: Credit to the answer here: https://stackoverflow.com/questions/62366377/how-to-type-generic-merging-function-in-typescript for the TypeScript magic.

Wozza365 (commented):
This is untestable against any of the examples as they are all plain JS. I can publish my project (which is TypeScript) to demo these changes if required.

src/kaboom.ts
slmjkdbtl: What about custom / user-defined component types here?

Wozza365: I wasn't aware of custom components being a thing - as I said in a comment elsewhere, this is my first time using Kaboom.

There are ways around it though, some better than others.

Something along the lines of:
`type AddParameters = (PosComp | ScaleComp | ... | Record<string, any>)`
Would allow for any object to be passed in. Most flexible - but also doesn't really offer the type definitions you'd want - for custom components this behaviour would be identical to the current live version.

Another option would be to redefine Comp's as interfaces to allow extending from a base interface. Something like:

`
interface IComp {
	___x___: boolean; // something is needed here for TS to enforce the interface, even if its an optional property. If it was all class based, it could be marked private and be less intrusive while being extendable.
}

export interface RotateComp extends IComp {
	angle: number,
}

// can be interface or class
class MyCustomComp extends IComp {
    myCustomProperty: number;
}

add<T extends IComp>(comps: ReadonlyArray<T>): Expand<UnionToIntersection<Defined<T>>> & GameObj

k.add([
   new MyCustomComp(),
   k.rotate(x, y)
])

`

This one takes better advantage of TypeScript, but is a more significant refactor. We could also do it with classes instead of interfaces which offers a little more flexibility with the problem in the base interface (which could be resolved if something sensible could live across all components?).

The latter would require anyone using TS to write a bit more code when defining a custom component. But I think that's a positive sacrifice for a strong type system (which is why people use TS in the first place).

I've got a lot more ideas on how the type system can be enhanced. We can allow for stuff like scene IDs to be strongly typed. I may open a couple more PRs in the coming days - some might be breaking changes if users have already rigged up their own types in weird ways :)

slmjkdbtl: Sorry for the late reply. Thank you! This works really well with interfaces. The other thing is, (forgive me I'm not a typescript expert), we're not actually assigning any `Comp` fields (they're all events / metadata) to the game obj, instead only the unique fields of those types that extends the `Comp` type, is there a way to "subtract" the `Comp` type fields?

slmjkdbtl: Thank you! It works wonderfully if change to `interface`, I'll merge.

PR #120, opened 2021-08-02
Update math.ts to use RGB values for clearColor for canvas bg color

zgrcnltnk:
Since WebGLRenderingContext.clearColor() does support r, g, b, a values only between 0 and 1, the RGB standard 0 - 255 doesn't work for current version. So, We need to use a simple normalization to turn RGB values into numbers between 0 and 1. I think this update was urgent to add, because choosing the right color between 0 and 1 isn't easy, and also, hex color support might be added as a feature in the next commits.

## Example:
 [58, 77, 178, 0.9] turns to [0.227, 0.302, 0.698, 0.9]


slmjkdbtl:
Hmm. I'd prefer it to be explicit that `rgb()` always take `0.0 - 1.0`, and you have to explicitly transform before you pass to `rgb()`, like `rgb(myColors.map((c) => c / 255.0))`

zgrcnltnk:
Yes, that one is another choice but, it needs extra conditions that one line will make the value false like if you normalize the already normalized value it will be always near to 0.

like [0.5, 0.25, 0.75, 1] -> [0,002, 0,0022, 0,0029, 1] and that'd cause show only white color for the canvas background.

As you said, another choice might be the place where clearColor is set with using rgb2Norm function obviously. Because you need to add some conditionals to make it right. I like short lines in code but, I think that one is a need. If you wish I can edit it that way, in the place where clearColor set.

slmjkdbtl:
I'm not sure if there's a tight detection solution here. E.g. `rgb2Norm()` will turn `(128, 1, 128)` into `(0.5, 1, 0.5)` which is wrong.

Is there a case where you're not sure if you're passing normalized rgb that would rely on detection?

zgrcnltnk:
At least, Phaser uses 0-255 range for the background parameters, and I think that would be much easier for the player who wants to code game fast :) If you wish you can merge it or take it the old way.

slmjkdbtl:
I'm down to consider changing everywhere in kaboom to 0-255, or a configuration setting that tells kaboom to use either 0.0-1.0 or 0-255, just not sure if having it inconsistent in different places is a good idea 🤔

zgrcnltnk:
Phaser uses rgb to 0-1 normalization. You can see that line starts in 231 at [Phaser Color.js](https://github.com/photonstorm/phaser/blob/v3.51.0/src/display/color/Color.js). And also there are redGL, blueGL, greenGL, alphaGL as normalized values of r, g, b values. See [Phaser Color API](https://photonstorm.github.io/phaser3-docs/Phaser.Display.Color.html)

slmjkdbtl:
Not merging this, but might do an option to have global 0-255 rgb in the future

PR #168, opened 2021-08-31
Add z index comp

### LINKED ISSUE #167: Not possible to have collisions and define render order
Right now it's only possible to define render order by either adding game objects after each other or specifying layers for them (and not have collisions)

It would be good to have a way to specify this, either by allowing to specify which layers can collide with each other or provide a way to define some sort of z coordinate to define the order of rendering.

finnp:
Objects on the same layer cann be ordered with `z(value: number)` comp.

Closes #167


slmjkdbtl:
Looks good, thanks!

PR #174, opened 2021-09-02
Change font size in rpg demo

lajbel:
The text looked very big and is not understood (and that my monitor is 1920 x 1080)

### DIFF
--- a/demo/rpg.js
+++ b/demo/rpg.js
@@ -104,7 +104,7 @@ scene("main", (levelIdx) => {
 
 	function talk(msg) {
 		talking = add([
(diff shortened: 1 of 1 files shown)


slmjkdbtl:
Can you get a screenshot? It's not visible on my machine for size 9 🤔 . Maybe screen pixel density difference?

slmjkdbtl:
Also I might change the default font to pixel font, I think the current default one it's a bit too stylized

lajbel:
image
With the default size

lajbel:
image
With 9 size

PR #185, opened 2021-09-03
Demo: Egg

lajbel:
I think it was obvious that such an example has to exist, **it is extremely necessary**, I suppose it contributes to the use of the `moveTo()` function without speed, and `use(sprite())`

Try in https://egg-demo.lajbel.repl.co/


slmjkdbtl:
This is important and needs to be merged immediately

slmjkdbtl:
I added a hatching process if you don't mind: https://kaboom2000.slmjkdbtl.repl.co/demo#egg

lajbel:
image
image
image

The miracle of life <3 (i love the egg desing, very good job tga)

PR #225, opened 2021-09-21
[RFC] Allow passing components directly as arguments, and support composition

masad-frost:
There's two folds for this RFC. Firstly, I think it would be good for beginners to not have to think about passing arrays to `add`. It should accept `add(body(), pos(), area())`. Secondly, as a side-effect of trying to be backwards incompatible, you can now compose components and use them in objects without having to learn array spread syntax. To fully support this form of composition, we can allow infinitely nested arrays.

Before

```ts
const playerComponents = [
       sprite('player'),
       body(),
       area(),
       pos(),
       "player",
];

const evilPlayerComponents = [
  ...playerComponents, // or spreading later when calling add([...playerComponents, ...evilPlayerComponents])
  color(0, 0, 0),
  origin("left"),
];

add(evilPlayerComponents);
```

After

```ts
const playerComponents = [
       sprite('player'),
       body(),
       area(),
       pos(),
       "player",
];

const evilPlayerComponents = [
  playerComponents, // or later when calling add(playerComponents, evilPlayerComponents)
  color(0, 0, 0),
  origin("left"),
];

add(evilPlayerComponents);
```

You can do `add([[[[[body()]]]], pos()])` or any form of nesting, it should still work!

The RFC is fully implemented and typescript supports it.

### DIFF
--- a/src/kaboom.ts
+++ b/src/kaboom.ts
@@ -334,7 +334,7 @@ const COMP_EVENTS = new Set([
 	"inspect",
 ]);
 
-function make<T>(comps: CompList<T>): Character<T> {
+function make<T>(comps: FlatCompList<T>): Character<T> {
 
 	const compStates = new Map();
 	const customState = {};
--- a/src/kaboom.ts
+++ b/src/kaboom.ts
@@ -527,11 +527,13 @@ function make<T>(comps: CompList<T>): Character<T> {
 
 }
 
-function add<T>(comps: CompList<T>): Character<T> {
-	const obj = make(comps);
+function add<A extends unknown[]>(...comps: [...A]): Character<RecursiveElementType<A>> {
+	const flatComps = comps.flat(Infinity) as RecursiveElementType<A>[];
+	const obj = make(flatComps);
 	obj._id = game.objs.push(obj);
 	obj.trigger("add");
 	ready(() => obj.trigger("load"));
+
 	return obj;
 }
 
--- a/src/types.ts
+++ b/src/types.ts
@@ -1423,6 +1423,11 @@ type Expand<T> = T extends infer U ? { [K in keyof U]: U[K] } : never;
 type MergeObj<T> = Expand<UnionToIntersection<Defined<T>>>;
 type MergeComps<T> = Omit<MergeObj<T>, keyof Comp>;
 
+// Gets an element type from a recursive array type
+// https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-1.html#recursive-conditional-types
+type RecursiveElementType<T> = T extends ReadonlyArray<infer U> ? RecursiveElementType<U> : T;
+
+
 type CompList<T> = Array<T | Tag>;
 type DynCompList<T> = CompList<T> | ((...args: any[]) => CompList<T>);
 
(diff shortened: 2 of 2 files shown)


slmjkdbtl:
I like this. Do you think we should generally advocate array-less `add()`? One thing is in `addLevel()` it [uses array](https://github.com/replit/kaboom/blob/master/demo/platformer.js#L109-L128) to define what each symbol means and calls `add()` later, wonder if we advocate array-less `add()` by default will it create more confusion when they're using `addLevel()`

masad-frost:
Yeah I saw that. Not sure to be honest. I can see people doing `'=': () => (sprit("poop"), solid())` which doesn't work.

My first thought for this was to have `compose` or `join` function and ditch arrays altogether i.e. `'=': () => compose(sprit("poop"), solid())`. But not sure how beginners would respond to it. It'd be good to do some user research here.

slmjkdbtl:
Another concern is if we should reserve the nested comps syntax. Kaboom (kinda) plans to moves to a node-tree structure where characters can have children who inherits transform attributes (position, scale, rotation), it'll be nice to use arrays to express this:

```js
// the parent
add([
    sprite("hero"),
    pos(100, 100),
    area(),
    // a child
    [
        sprite("sword"),
        pos(-10, 0),
    ],
    // another child
    [
        sprite("pet"),
        float(),
    ],
]);
```
but not sure yet.

masad-frost:
Also thinking about

```
hero = add([
    sprite("hero"),
    pos(100, 100),
    area(),
    parent(), // gives addChild capabilities
]);

hero.addChild([ // same API as `add`
    sprite("sword"),
    pos(-10, 0),
]);
```

slmjkdbtl:
Yep, or maybe
```js
const sword = add(...);
const pet = add(...);

add([
    sprite("hero",
    children(sword, pet), 
]);
```

I'll merge this to allow the syntax, but will think more if we want to advocate this syntax over the existing one.

src/kaboom.ts
```diff
@@ -527,11 +527,13 @@ function make<T>(comps: CompList<T>): Character<T> {
 
 }
 
-function add<T>(comps: CompList<T>): Character<T> {
-	const obj = make(comps);
+function add<A extends unknown[]>(...comps: [...A]): Character<RecursiveElementType<A>> {
```
slmjkdbtl, line 530: nit: do you think if this will scare people on autocomplete? is there a way to alias stuff into an easier looking format

masad-frost: Yes... But if I'm reading things correctly, as it stands this function's signature is not exposed for autocomplete, only signature exposed is the one in types.ts, for better or worse.

slmjkdbtl: Hmm but if it's not exposed in types.ts the new syntax won't resolve in the type check

slmjkdbtl: (the `kaboom.d.ts` that's exposed is generated only from `types.ts`)

masad-frost: Yeah true. Lemme think about it.

slmjkdbtl:
also i think the blockly version of kaboom (which targets more beginners) will be hiding the array syntax for `add()` anyway

PR #234, opened 2021-09-24
Added support to x-browser fullscreen

### LINKED ISSUE #213: Error: app.canvas.requestFullscreen is not a function, when call fullscreen()
## Description
It occurs when I use fullscreen() in some browsers like Safari, both on MacOS and iOS.
Seems the Fullscreen API in some browser engines should have prefix, such as `webkit`, `moz`, `ms`.
If you need, I could create a PR for it. 

## References:
- [Fullscreen_API](https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API)
- [sindresorhus/screenfull.js](https://github.com/sindresorhus/screenfull.js)

mannysz:
Fixes #213

Added the cross browser calls in the fullscreen function defined in app.ts.
Caveats:
- It works on Google Chrome
- It works on Safari Experimental Build for OS X
- Does not work on current Safari stable version
- Does not work on last iOS Safari stable version
- For mobile version, I recommend using Capacitor or Cordova and packing a native app.

Needs help with:
- Testing in Mozilla Firefox
- Testing with MS Edge

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -361,12 +361,39 @@ function appInit(gconf: AppConf = {}): App {
 	}
 
 	function fullscreen(f?: boolean): boolean {
-		if (document.fullscreenElement) {
-			document.exitFullscreen();
+		const enterFullscreen = (el: any) => {
+			if(el.mozRequestFullScreen) {
+				el.mozRequestFullScreen();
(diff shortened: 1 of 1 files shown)


slmjkdbtl:
Thank you! Looks good

PR #245, opened 2021-09-25
add multiplayer entry for template

slmjkdbtl:
Enable people to just edit `multiplayer.js` to program multiplayer logic with web sockets.


sergeichestakov (approved):
🧑‍🤝‍🧑

Issue #246, opened 2021-09-26
Add dark mode to Kaboom docs


techpixel:
It would be nice if the Kaboom docs would adjust to the system default mode (dark or light) when browsing the docs.

techpixel:
Also I really liked this feature, why was it removed?
![image](https://user-images.githubusercontent.com/68567672/134838341-2ed12195-5179-49fc-984c-467a26160c0c.png)

slmjkdbtl:
It's not! Just WIP. I'm just working on dark theme, I'll also add this back.

techpixel:
Also: is there a discord support server?

slmjkdbtl:
[Yep](https://discord.gg/aQ6RuQm3TF)

slmjkdbtl:
Dark mode is added

techpixel:
Thank you! Also, could you add dark mode to the landing page?

PR #247, opened 2021-09-27
[site] add dark mode

slmjkdbtl:
- tokenized css color
- added button in doc and demo page

https://user-images.githubusercontent.com/9929523/134844490-9600dc6e-97e0-45f9-8ae7-b1a4ddc0add5.mp4

https://user-images.githubusercontent.com/9929523/134844499-06da2813-231a-4e7e-885b-5ea9d573c1d1.mp4


slmjkdbtl:
#246

sergeichestakov (approved):
Ooh neat! 🌃 

I think it's worth checking system preferences and defaulting to that if it's set (e.g. https://stackoverflow.com/questions/50840168/how-to-detect-if-the-os-is-in-dark-mode-in-browsers).

Also, perhaps this should be an actual toggle rather than a button?

slmjkdbtl:
That's on my mind too! Just pushed those

PR #250, opened 2021-09-28
rename clearColor to background

slmjkdbtl:
`clearColor` is more of word from webgl, not as intuitive as the word `background`, confused some users.


sergeichestakov (approved):
🖼️

PR #253, opened 2021-09-28
move site to nextjs

slmjkdbtl:
The website might grow, nextjs is more extendible than the previous system which is mainly for very static pages.


demc:
@slmjkdbtl This looks cool! What kind of feedback are you looking for? Your PR description was a little bit too succinct to tell what you're planning to do!

demc:
Please reboop for review after adding extra context!

PR #260, opened 2021-10-05
fix collides() checking for the same obj

slmjkdbtl:
- fixed a typo that's causing objects to check collisions with themselves
- publishes the fix


sergeichestakov (approved):
Nice fix! 🔨 In general tho, think it's worth keeping publishing in separate commits

slmjkdbtl:
Sounds good 👍  Will do

PR #277, opened 2021-10-12
Expose more collision detection functions

slmjkdbtl:
There are some cases where you can't totally depend on `area()` and have to do some manual collision checking, useful to expose some collision detection pure functions.

### DIFF
--- a/src/kaboom.ts
+++ b/src/kaboom.ts
@@ -15,13 +15,12 @@ import {
 	mapc,
 	wave,
 	testLineLine,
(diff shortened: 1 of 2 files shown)


mikewesthad:
This seems reasonable to me! I do wonder about the naming. In a game library, I typically expect something with collide in the name to do some kind of collision resolution (or at least return some physics-y collision data like displacement to resolve the collision). Would these be better as something like `intersectLineVsLine` or `testLineVsLine`?

PR #280, opened 2021-10-13
Env all folders in static 

lajbel:
it will make it easier for new or inexperienced users to add more folders like a fonts folder or comps folder, since they will not have to edit the run.js file to be able to use their files from their new folders


slmjkdbtl:
Makes sense 👍 Thanks

PR #289, opened 2021-10-16
Don't run loop when document not visible

slmjkdbtl:
Visibility change on different browsers are weird, sometimes causing high delta time surge which for some reason causes crash, this is a hot fix to not run the game loop at all when document is not visible, eliminating some delta time surges.


masad-frost:
essentially pausing the game?

PR #298, opened 2021-10-18
Fix fullscreen() behavior

slmjkdbtl:
`fullscreen(true)` should go into fullscreen, `fullscreen(false)` should exit, and `fullscreen()` should just return the current fullscreen state. Currently it always go into fullscreen

- fixed the logic
- cleaned up a bit
- added some warnings about stuff that's not working atm in the doc


slmjkdbtl:
That makes sense 👍

PR #310, opened 2021-10-19
[burp] Add burp mode

slmjkdbtl:
Popular demand


src/kaboom.ts
```diff
@@ -777,6 +777,10 @@ function regDebugInput() {
 
 }
 
+function enterBurpMode() {
+	keyPress("b", audio.burp);
```
DaInfLoop, line 781: Absolutely beautiful, and yes I'm late

PR #318, opened 2021-10-20
burp with burp

lajbel:
haha funny title.


slmjkdbtl:
i'd remove `keyPress()` and use burp mode but keep the `mouseClick` one, so people can have burp fun on their mobile

PR #328, opened 2021-10-21
Add visual novel demo 

lajbel:


slmjkdbtl:
also `tw-` is my initials hh, we use it at replit to tell who owns which branch

lajbel:
let me fix a couple of things

PR #373, opened 2021-10-29
Support all mouse buttons

slmjkdbtl:
- Support adding the first argument to `onMouseXXX` to specify which button to listen to
```js
onMousePress("right", showMyMenu)
```
- Renamed `onMouseClick` to `onMousePress` to share a similar naming style like `onKeyPress`, support `onClick` without the first `tag` argument to catch all moue click events. Removed `onMouseClick` in favor of `onMousePress` or `onClick`
```js
// before
onMouseClick(dragPlayer);
// after
onClick(dragPlayer);
```

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -4,6 +4,7 @@ import {
 
 import {
 	Key,
+	MouseButton,
 	Vec2,
 	Cursor,
 } from "./types";
@@ -32,7 +33,7 @@ type AppCtx = {
 	canvas: HTMLCanvasElement,
 	mousePos: Vec2,
 	mouseDeltaPos: Vec2,
-	mouseState: ButtonState,
+	mouseStates: Record<string, ButtonState>,
 	keyStates: Record<string, ButtonState>,
 	charInputted: string[],
 	mouseMoved: boolean,
@@ -59,9 +60,9 @@ type App = {
 	keyPressed(k?: Key): boolean,
 	keyPressedRep(k?: Key): boolean,
 	keyReleased(k?: Key): boolean,
-	mouseDown(): boolean,
-	mouseClicked(): boolean,
-	mouseReleased(): boolean,
+	mouseDown(m?: MouseButton): boolean,
+	mousePressed(m?: MouseButton): boolean,
+	mouseReleased(m?: MouseButton): boolean,
 	mouseMoved(): boolean,
 	charInputted(): string[],
 	cursor(c?: Cursor): Cursor,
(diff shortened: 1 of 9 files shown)


src/app.ts
```diff
@@ -281,12 +297,16 @@ function appInit(gopt: AppOpt = {}): App {
...
 
 	app.canvas.addEventListener("touchcancel", (e) => {
 		if (!gopt.touchToMouse) return;
-		app.mouseState = "released";
+		app.mouseStates["left"] = "released";
+	});
+
+	app.canvas.addEventListener("contextmenu", function (e) {
+		e.preventDefault();
 	});
```
slmjkdbtl, line 310: kinda unrelated, but context menu doesn't make much sense in a canvas, and we support customizing right mouse button now

src/app.ts
```diff
@@ -319,16 +339,16 @@ function appInit(gopt: AppOpt = {}): App {
 		return app.mouseDeltaPos.clone();
 	}
 
-	function isMouseClicked(): boolean {
```
slmjkdbtl, line 322: Should this default to left mouse button, or any mouse button like how `isKeyPressed()` without argument works?

src/kaboom.ts
```diff
@@ -688,13 +689,17 @@ function onCollide(
...
+		return onMousePress(tag);
+	} else {
+		return onUpdate(tag, (o: GameObj) => {
+			if (!o.area) throw new Error("onClick() requires the object to have area() component");
+			if (o.isClicked()) {
+				cb(o);
+			}
+		});
+	}
 }
```
slmjkdbtl, line 703: Function overloading is ugly to implement in JS... Any tips?

src/app.ts
```diff
@@ -319,16 +339,16 @@ function appInit(gopt: AppOpt = {}): App {
...
-		return app.mouseState === "pressed";
+	function isMousePressed(m = "left"): boolean {
```
slmjkdbtl, line 342: If you don't pass an argument to isKeyPressed(), it'll return true if any key was just pressed. Wonder if why should do the same here to return true if any mouse button was just pressed, or default to left?

sergeichestakov: sounds like a bit of an edge case but why not just make this param required? seems confusing otherwise

sergeichestakov (approved):
Looks good just one suggestion

src/app.ts
slmjkdbtl: The assumption is people mainly deal with the left mouse button, maybe this can save some typing. Or maybe people will assume this function would check for left mouse by default?

PR #376, opened 2021-10-30
Add more educational demos

slmjkdbtl:
- Added some more educational demos under 
- Removed some redundant or not working demos
- Tweaked the demo dropdown on `/play` to have some order (put the most intro level demos on top)


sergeichestakov (approved):
🦘💥

PR #381, opened 2021-11-01
Add color constants

slmjkdbtl:
Nice to have?

### DIFF
--- a/src/kaboom.ts
+++ b/src/kaboom.ts
@@ -2702,6 +2702,15 @@ const ctx: KaboomCtx = {
 	RIGHT: vec2(1, 0),
 	UP: vec2(0, -1),
 	DOWN: vec2(0, 1),
+	// colors
+	RED: rgb(255, 0, 0),
+	GREEN: rgb(0, 255, 0),
+	BLUE: rgb(0, 0, 255),
+	YELLOW: rgb(255, 255, 0),
+	MAGENTA: rgb(255, 0, 255),
+	CYAN: rgb(0, 255, 255),
+	WHITE: rgb(255, 255, 255),
+	BLACK: rgb(0, 0, 0),
 	// dom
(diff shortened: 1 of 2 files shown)


src/kaboom.ts
```diff
@@ -2702,6 +2702,15 @@ const ctx: KaboomCtx = {
...
 	DOWN: vec2(0, 1),
+	// colors
+	RED: rgb(255, 0, 0),
+	GREEN: rgb(0, 255, 0),
+	BLUE: rgb(0, 0, 255),
+	YELLOW: rgb(255, 255, 0),
+	MAGENTA: rgb(255, 0, 255),
+	CYAN: rgb(0, 255, 255),
+	WHITE: rgb(255, 255, 255),
+	BLACK: rgb(0, 0, 0),
```
sergeichestakov, line 2713: hmm perhaps we should namespace this under `colors` (e.g. `kaboom.colors.RED`) - same with dirs above (e.g. `kaboom.dirs.LEFT`) although that'll be a breaking change if this is out already.

sergeichestakov (approved):
Looks good - I would recommend namespacing it though so that the top level constant doesn't get too crowded

src/kaboom.ts
slmjkdbtl: Yeah the direction vectors are already global and out. Don't have strong opinion on this, but one concern is global might be friendlier to use especially in kaboom context a lot of times you're putting them in the `color()` function
```js
add([
    rect(40, 20),
    color(colors.RED),
])
```
```js
add([
    rect(40, 20),
    color(RED),
])
```

slmjkdbtl:
Going to merge this version for easiness to write and consistency with the already top level direction constants, but interested to revisit and possibly make all constants namespaced on next breaking release

PR #383, opened 2021-11-02
Don't try to stretch canvas if it's provided by the user

slmjkdbtl:
Currently we're stretching all canvas to parent size if `width` and `height` are not given, but it doesn't make sense to do that for user provided canvas DOM since they might already setup their ideal canvas size with html or css.


moudy (approved):
sounds reasonable. but do you think people could be relying on the previous behavior?

slmjkdbtl:
Hmm, I think there is a chance. I'll think more on this

PR #384, opened 2021-11-02
Expose debug.curRecording

slmjkdbtl:
F6 to record is managed by kaboom debug mode, it makes sense to expose the recording handle under `debug`

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -3368,6 +3368,10 @@ export interface Debug {
 	 * Log an error message to on screen debug log.
 	 */
 	error(msg: string): void,
+	/**
+	 * The recording handle if currently in recording mode.
+	 */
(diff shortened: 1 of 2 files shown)


src/types.ts
```diff
@@ -3368,6 +3368,12 @@ export interface Debug {
 	 * Log an error message to on screen debug log.
 	 */
 	error(msg: string): void,
+	/**
+	 * The recording handle if currently in recording mode.
+	 *
+	 * @since v2000.1.0
```
moudy, line 3374: haha. i though this was a date at first

moudy (approved):

src/types.ts
slmjkdbtl: hh 😆

Issue #386, opened 2021-11-03
Suggestion: Enable multiple tags


oaklandgit:
It would be useful to be able to target multiple tags at once by passing in an array of tags wherever tags can be used:
```
k.every(["enemy","bullet"], (obj) => {
  obj.paused = true
})
```
Or something like that.

slmjkdbtl:
Implemented in #387 !

oaklandgit:
Awesome! Thank you @slmjkdbtl

PR #387, opened 2021-11-03
Support multiple tags on get() and every()

slmjkdbtl:
#386

`GameObj#is()` always supported checking for multiple tags, it makes sense to support multiple tags for `get()` and `every()` too


sergeichestakov (approved):
Neat! 🏷️

PR #394, opened 2021-11-05
Expose formatText() and drawFormattedText()

slmjkdbtl:
The main use case is to get the size of the text without rendering it. It's used internally in various places to draw text backgrounds, nice to expose.

And renamed from previously internal `fmtText()` to `formatText()`, `drawFmtText()` to `drawFormattedText()`

### DIFF
--- a/src/kaboom.ts
+++ b/src/kaboom.ts
@@ -248,13 +248,35 @@ function drawSprite(opt: DrawSpriteOpt) {
 
 // wrapper around gfx.drawText to integrate with font assets mananger / default font
 function drawText(opt: DrawTextOpt) {
-	// @ts-ignore
-	const fid = opt.font ?? DEF_FONT;
-	const font = assets.fonts[fid];
+	const font = (() => {
+		if (opt.font) {
+			return typeof opt.font === "string" ? assets.fonts[opt.font] : opt.font;
+		} else {
+			return assets.fonts[DEF_FONT];
+		}
+	})();
 	if (!font) {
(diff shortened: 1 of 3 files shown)


src/kaboom.ts
```diff
@@ -248,18 +248,40 @@ function drawSprite(opt: DrawSpriteOpt) {
 
 // wrapper around gfx.drawText to integrate with font assets mananger / default font
 function drawText(opt: DrawTextOpt) {
```
slmjkdbtl, line 250: unrelated, cleaning up the logic to find the font

src/types.ts
```diff
@@ -2770,7 +2808,7 @@ export type DrawTextOpt = RenderProps & {
 	/**
 	 * The name of font to use.
 	 */
-	font?: string,
+	font?: string | FontData,
```
slmjkdbtl, line 2811: unrelated, but nice to also enable passing in raw font data from `loadFont`, also makes the type more compatible with the internal version and the exposed version

src/kaboom.ts
```diff
@@ -220,18 +220,20 @@ function mouseWorldPos(): Vec2 {
 	return game.camMousePos;
 }
 
+function findAsset<T>(src: string | T, lib: Record<string, T>, def?: string): T | undefined {
```
slmjkdbtl, line 223: unrelated, cleaning up logic to find asset since it's coming up a lot

sergeichestakov (approved):
Neat! ✏️

PR #402, opened 2021-11-09
Add testRectCircle

b5710546232:
Hi I would like to help add testRectCircle :)

It is checked by
 1 find the nearest point with circle
 2 find the distant between the nearest point 
 3 check the distant if it is less than circle's radius then it is intersect


slmjkdbtl:
Nice! Thank you

slmjkdbtl (approved):

PR #405, opened 2021-11-10
Add Isometric level support

gugiserman:
Hey guys,

I just recently found out about `kaboom` and I was totally in love with it, but I really wanted to be able to work with isometric tilemaps (mostly generated via [Tiled](https://www.mapeditor.org/)). I was going to work on it as a plugin or maybe fork already existing plugins but I felt like this should be supported by Kaboom first and then integrated with external software as add-ons for extended support and more advanced use cases.

New logic added is mostly `addIsometricLevel` function in `/src/kaboom.ts`

---

image

---

Obs: I thought of maybe adding a `isometric?: boolean` to `addLevel` and make conditional rendering from there, but decided to go with separate method instead, so let me know what you think!


gugiserman:
Noticed some obvious mistakes I made and fixed them but its still half diamond shape, I'm not sure why:

image

gugiserman:
Nevermind, I was tripping. I think it's all working now, just need to test the offset thing.

image

slmjkdbtl:
Thank you! This will be an exciting feature

src/types.ts
```diff
@@ -3806,6 +3806,27 @@ export interface LevelOpt {
 	[sym: string]: any,
 }
 
+export interface IsometricLevelOpt {
+	/**
+	 * Width of each block.
+	 */
+	tileWidth: number,
```
slmjkdbtl, line 3813: Do you think this can be `width` and `height` to be consistent with `addLevel`? I do think `tileWidth` might be a better name, maybe we can deprecated `width` on both places and move to a more explicit name

gugiserman: @slmjkdbtl yeah definitely! I did that just to make it easier for me to grasp about the code since it was my first time viewing the codebase. I will mirror property names from `addLevel` right now.

src/kaboom.ts
```diff
@@ -2458,6 +2472,109 @@ function addLevel(map: string[], opt: LevelOpt): Level {
...
+
+		gridHeight() {
+			return options.height;
+		},
+
+		fromIsometricCoordsToWorldPos: (row: number, col: number): Vec2 => {
+			return vec2((col - row) * halfTileWidth, (col + row) * halfTileHeight);
+		},
+
+		from2dTo1dIndex: (row: number, col: number): number => {
```
slmjkdbtl, line 2506: When would this function be useful to user (there doesn't seem to be a way to index from this 1d index)?

slmjkdbtl:
I'm not familiar with isometric games, is it common to make sprites directly in isometric format, or is it common to make sprites in normal format, but use skew transform to make it look isometric (skew transform doesn't exist in kaboom yet but is trivial to implement)?

src/kaboom.ts
gugiserman: I thought this could enable the user to easily lookup the original level array to get the symbol for a specific tile, and from the symbol they can get its metadata in another object or something.

Is there a way to get which tile the user clicked on for example? I thought of the following use case but I'm not familiar with kaboomjs enough to say more yet:

1- player is sitting at tile X,Y
2- player clicks on tile W,Z
3- from W,Z tile coord you can get one dimensional array position (k) to lookup level declaration and get tile symbol
4- from tile symbol you can get some info elsewhere or just run some kind of pathfinding algorithm from X,Y (i) to W,Z (k).
5- walk along the generated path considering collision etc

Is there a better way to implement something like A* pathfinding algorithm in this case?

Sorry I haven't done much in kaboom yet so I'm not so familiar with the API yet, I'm waiting for the weekend to come x)

gugiserman:
> I'm not familiar with isometric games, is it common to make sprites directly in isometric format, or is it common to make sprites in normal format, but use skew transform to make it look isometric (skew transform doesn't exist in kaboom yet but is trivial to implement)?

It is ideal to make it in isometric format from start, but there are tools and techniques to convert traditional topdown squared assets like terrain tiles into isometric formats. Sometimes the outcome is perfect and seems like it was planned and done in isometric, sometimes not so much...

For characters and other objects I think there's usually no way around, unlike terrain tilesets you can't really play around with resizing and skewing it and expect it to look good enough 😝 

You can take a look at various different isometric resources, from flat terrain tiles to block ones. I'm not sure how we would go about supporting the block tile type though.

slmjkdbtl (approved):

src/kaboom.ts
```diff
@@ -2458,6 +2472,109 @@ function addLevel(map: string[], opt: LevelOpt): Level {
...
+
+		gridWidth() {
+			return options.width;
+		},
+
+		gridHeight() {
+			return options.height;
+		},
+
+		fromIsometricCoordsToWorldPos: (row: number, col: number): Vec2 => {
```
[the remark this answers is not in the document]
gugiserman: @slmjkdbtl sure I will rename it to `toWorldPos()`

gugiserman:
@slmjkdbtl any idea what could have happened here? I simply rebased the branch and adjusted `getPos()` of `isometricLevel` to match `level`'s interface.

image

slmjkdbtl:
Weird, maybe the break is introduced by 72b59c8ce50c7369d9ef52d211901fe48e299ff6 which does some refactoring on how game objects work, can you try if reverting to 
 rebase on the previous commit 03c23a292a8792be6e238c37d77b731e481486b2 to see if that's the case

gugiserman:
> Weird, maybe the break is introduced by [72b59c8](https://github.com/replit/kaboom/commit/72b59c8ce50c7369d9ef52d211901fe48e299ff6) which does some refactoring on how game objects work, can you try if reverting to rebase on the previous commit [03c23a2](https://github.com/replit/kaboom/commit/03c23a292a8792be6e238c37d77b731e481486b2) to see if that's the case

Yep, it works on top of https://github.com/replit/kaboom/commit/03c23a292a8792be6e238c37d77b731e481486b2

image

slmjkdbtl:
Thanks 👍 I'll investigate the break

slmjkdbtl:
Sorry for the delay. The issue should be fixed in #468, if it does feel free to merge anytime

gugiserman:
> Sorry for the delay. The issue should be fixed in #468, if it does feel free to merge anytime

Thanks! But it's still pending your review before I can merge.

slmjkdbtl (approved):

slmjkdbtl:
Sorry thought already approved 😅

gugiserman:
@slmjkdbtl hello man, I just ran a very difficult rebase of my branch and was checking if everything is good. Can you help me out here if there's still any chance of us merging this branch? 😆 

I'm getting this one error from the newly added demo when running the test suite. I assume there's a list of globals this needs to be included, but I can't seem to find it. A lot of changes since I first created this PR which is awesome 👍🏻 

image

I have also noticed `addLevel` now returns a `GameObj`, but `addIsometricLevel` still returns a `IsometricLevel`. Let me know if that needs to be refactored to mirror `addLevel` or not really.

Other than that, it seems like everything else is in order and it still works.

image

jumpjack:
Does it exist a tutorial or demo for isometric games with kaboom?

slmjkdbtl:
Hey, sorry for very much delay on this PR. Was away from kaboom for a while and recently got back.

The level system is going to go under some big change soon so I'm thinking to merge this after the change is done so we only have to rebase to the new code once.

PR #408, opened 2021-11-10
enterState check fail fix

micheg:
The states are an array of strings, but the comparison is done as if it were a dictionary.
I believe the code should be: if (stateList && !stateList.includes(state))
which seems more sensible to me.

### DIFF
--- a/src/kaboom.ts
+++ b/src/kaboom.ts
@@ -2153,7 +2153,7 @@ function state(initState: string, stateList?: string[]): StateComp {
 		id: "state",
 		state: initState,
 		enterState(state: string, ...args) {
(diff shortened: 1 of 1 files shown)


slmjkdbtl (approved):
Oops, that's a big bug. Thanks for the fix!

micheg:
np :-)
I am playing a bit with the library in the evening after work, for
religious reasons I do not look at the code if it does not strictly follow
the Allman style :-) But this time I made an exception :-)

Il giorno mer 10 nov 2021 alle ore 18:53 tga ***@***.***> ha
scritto:

slmjkdbtl:
Looks like you closed before merge?

PR #410, opened 2021-11-11
Node tree for game objects

slmjkdbtl:
## Overview

This PR extends the existing flat-list style game object into a node tree structure, each game object can have a list of children.

- Nested transformation (position, scale, rotation)
- Group visibility / activity toggle

```js
const player = add([
    sprite("froggy"),
    pos(100, 20),
    area(),
])

// sword will follow player position with offset vec2(12)
const sword = player.add([
    sprite("sword"),
    area(),
    pos(12),
    rotate(45),
    weapon(),
])

onKeyPress("space", () => {
    sword.swing()
})

onKeyPress("h", () => {
    // both player and sword will go hidden
    player.hidden = true
})

sword.parent // => player
player.children // => [ sword, ]
```

- Deprecate layers, use parent game objects as layer

```js
const ui = add([
    z(100),
    fixed(),
])

ui.add([
    text("score: " + score),
    pos(24),
])
```
also gives better flexibility to do stuff like parallax scrolling

- Simpler internals. Everything is a game object, kaboom manages a root game object `root`, the exported `add()` is an alias of `root.add()`

## Open Questions

Should these properties also inherit from parent:
- color
- opacity
- shader


oaklandgit:
This is awesome. 
Question: if layer is deprecated in favor of parents (e.g. `ui.add([etc.])` ) how would a symbol added to a level have that parent gameObject associated with it?

Example:

```js
"@" : () => [
  // how to declare its parent?
]
```

slmjkdbtl:
`Level` is planned to be transformed into a game object and be the parent of all tiles, and you can use `z()` to define the order of that whole level.

oaklandgit:
@slmjkdbtl Curious about the status of scene graph. Last I saw collisions weren't working?

slmjkdbtl:
Collision detections are working on master now (haven't released an alpha version for that yet), but the physics are still a bit glitchy

oaklandgit:
Fantastic. I'll give it a try!

PR #418, opened 2021-11-15
Fix missing state event not initilized in some cases

slmjkdbtl:
Currently `state()` component will always report error if you don't bind an `onStateUpdate()` event (the `update()` hook will try to access the event handlers), due to the state events are not initialized unless you bind an event.

This PR wraps the event trigger and register in `on()` and `trigger()` helpers to make sure there's always initialization before usage.


sergeichestakov (approved):
🔨

PR #422, opened 2021-11-16
(feature) Implement temporary culling option in cleanup comp

gugiserman:
I'm not entirely sure if this one is desirable since kaboom don't seem to aim to be a solution for massive game worlds etc, but I needed this in my project and seemed easy enough to achieve, so...

### DIFF
--- a/src/kaboom.ts
+++ b/src/kaboom.ts
@@ -1278,7 +1278,7 @@ function move(direction: number | Vec2, speed: number): MoveComp {
 	};
 }
 
-function cleanup(time: number = 0): CleanupComp {
+function cleanup(time: number = 0, cull: boolean = false, cullOffset: Vec2 = vec2(0, 0)): CleanupComp {
 	let timer = 0;
 	return {
 		id: "cleanup",
@@ -1288,12 +1288,27 @@ function cleanup(time: number = 0): CleanupComp {
 				p1: vec2(0, 0),
 				p2: vec2(width(), height()),
 			}
+
+			if (cull) {
+				screenRect.p1.x -= cullOffset.x
+				screenRect.p1.y -= cullOffset.y
+				screenRect.p2.x += cullOffset.x
+				screenRect.p2.y += cullOffset.y
+			}
+
 			if (testAreaRect(this.screenArea(), screenRect)) {
 				timer = 0;
+				if (cull) {
+					this.hidden = false
+				}
 			} else {
 				timer += dt();
 				if (timer >= time) {
-					this.destroy();
+					if (cull) {
+						this.hidden = true
+					} else {
+						this.destroy();
+					}
 				}
 			}
 		},
--- a/src/types.ts
+++ b/src/types.ts
@@ -450,7 +450,7 @@ export interface KaboomCtx {
 	 * ])
 	 * ```
 	 */
-	cleanup(time?: number): CleanupComp,
+	cleanup(time?: number, cull?: boolean, cullOffset?: Vec2): CleanupComp,
 	/**
 	 * Follow another game obj's position.
 	 */


slmjkdbtl:
Yeah ideally we can do this by default for everything in a cheap way. One idea was to check if every final transformed screen space vertex is within [0.0 - 1.0] [here](https://github.com/replit/kaboom/blob/9d19cf7fd0360867ec9ee7278791d877c0a82177/src/gfx.ts#L483), but I'm not sure if it'll be more expensive if the user rarely have out-of-screen stuff (can be an opt-in feature with a flag in `kaboom()`)

Do you have any data on how much performance this can save?

gugiserman:
> Do you have any data on how much performance this can save?
> 
> Yeah ideally we can do this by default for everything in a cheap way. One idea was to check if every final transformed screen space vertex is within [0.0 - 1.0] [here](https://github.com/replit/kaboom/blob/9d19cf7fd0360867ec9ee7278791d877c0a82177/src/gfx.ts#L483), but I'm not sure if it'll be more expensive if the user rarely have out-of-screen stuff (can be an opt-in feature with a flag in `kaboom()`)

I don't, but I can gather some later. I had to do it because my level was 71x71 with fairly big resolution and screen was stutterring a lot on character move. This solved my issue for now at least.

slmjkdbtl:
@gugiserman So it did obviously help with the framerate in your case?

slmjkdbtl:
Can you try if #426 achieve similar or better performance?

gugiserman:
> Can you try if #426 achieve similar or better performance?

I will try them both later today. Thanks 😄

gugiserman:
@slmjkdbtl  Quick tests with bean and a isometric level from my other PR just as a sample of my real project:

### Recordings from mac native screen recorder:

No `cull` option nor `cleanup` comp:
https://user-images.githubusercontent.com/4623123/142053513-7dc38102-ce89-4ef4-84aa-0ae2e7e3e54a.mov

`cull` kaboom option:
https://user-images.githubusercontent.com/4623123/142053653-caa28e02-5ccd-4a2a-b2b1-c7a54e7ce115.mov

`cleanup` comp without `cullOffset`
https://user-images.githubusercontent.com/4623123/142054218-f00128a1-8456-4ad0-a0aa-0314f31cdf7d.mov

`cleanup` comp with 400x400 `cullOffset`
https://user-images.githubusercontent.com/4623123/142053840-a4ad1528-b5e4-452b-833d-c64262ff5494.mov

`cleanup` comp with 6x6 tiles worth of `cullOffset`
https://user-images.githubusercontent.com/4623123/142053985-3f2ff011-dc0a-4482-8bf1-86bfeec96aff.mov

### Recordings using kaboom's `record()`:

No `cull` option nor `cleanup` comp:
https://user-images.githubusercontent.com/4623123/142056195-2cab1ca0-b5de-4a8d-9f22-914cd70ef320.mp4

`cull` kaboom option:
https://user-images.githubusercontent.com/4623123/142056248-c840ce42-57c9-4764-8c65-c893e1e9d65d.mp4

`cleanup` comp without `cullOffset`
https://user-images.githubusercontent.com/4623123/142056266-65970685-1686-496f-a380-e4d16abdfebf.mp4

`cleanup` comp with 400x400 `cullOffset`
https://user-images.githubusercontent.com/4623123/142056281-1283ffd6-c8fb-424b-92c5-75188952f248.mp4

`cleanup` comp with 6x6 tiles worth of `cullOffset`
https://user-images.githubusercontent.com/4623123/142056309-65863838-ac1b-47fc-af42-48fcd5e9774d.mp4

slmjkdbtl:
Thanks for the tests! 

One thing I'm not sure is if this should be a separate component instead of a feature in `cleanup()`, since culling is not necessarily cleaning anything up.

What about a separate component with a more expressive name that handles general out of view logic

```js
add([
    sprite("froggy"),
    outOfView({
        pause: true,
        hidden: true,
        destroy: false,
        onOut: () => debug.log("oh no"),
    }),
])

// cleanup() can be converged to this component with pre config 
const cleanup = () => outOfView({ destroy: true })
```

(`outOfView` is slightly verbose comparing to other built-in component names, any other ideas?)

slmjkdbtl:
One thing I'm not sure is if this should be a separate component instead of a feature in `cleanup()`, since culling is not necessarily cleaning anything up.

What about a separate component with a more expressive name that handles general out of view logic

```js
add([
    sprite("froggy"),
    outOfView({
        pause: true,
        hidden: true,
        destroy: false,
        onOut: () => debug.log("oh no"),
    }),
])

// cleanup() can be converged to this component with pre config 
const cleanup = () => outOfView({ destroy: true })
```

(`outOfView` is slightly verbose comparing to other built-in component names, any other ideas?)

gugiserman:
Yeah definitely a separate component makes sense. Hard one to name though. `outOfView` is intuitive enough I guess. I like the config approach, very versatile.

gugiserman:
> Thanks for the tests! However I'm not fully sure what's the performance on each one, do you have data on what FPS is on which option?

I can do that some other day and maybe put some fps charts for each one, but I don't know how else to measure performance on something like this.

slmjkdbtl:
Looks like it's kinda hidden in the docs, but you can use `debug.fps()` to get the current fps, or use `debug.inspect = true` or press `F1` to enter inspect mode which shows the fps at top left

PR #426, opened 2021-11-16
Option to not render objects outside of screen

slmjkdbtl:
With option

```js
kaboom({ cull: true })
```

you can make kaboom don't render objects outside of screen. Can help with performance in large worlds where there's a ton of stuff out screen.

It'll be nice if it's enabled by default, however it does do a lot of additional calculation to check if the object is totally out screen which is unnecessary for on-screen objects that's definitely going to be rendered. On my device turning on cull doesn't seem to affect the performance of on-screen rendering, but it's a fairly powerful CPU and not sure about lower-end devices yet.


masad-frost (approved):
Sweet!

I wonder if it's worth making into a component for objects as well. That way you can only apply it to objects that you know might go off screen.

masad-frost:
> but it's a fairly powerful CPU and not sure about lower-end devices yet.

You can throttle CPU from chrome devtools  
image

slmjkdbtl:
Was just writing about [a general out of view component](https://github.com/replit/kaboom/pull/422#issuecomment-970697493)

PR #427, opened 2021-11-16
Add transition guard to state()

slmjkdbtl:
As proposed in [this comment](https://github.com/replit/kaboom/pull/346#pullrequestreview-789795171)

```js
const s = state("move", [ "idle", "attack", "move", ], {
    "idle": [ "move", "attack" ],
    "attack": "idle",
    "move": "attack",
})

// define event for specific transitions
s.onStateTransition("move", "attack", throwFeces)
```

### DIFF
--- a/src/kaboom.ts
+++ b/src/kaboom.ts
@@ -2133,7 +2133,7 @@ function lifespan(time: number, opt: LifespanCompOpt = {}): LifespanComp {
 function state(
 	initState: string,
 	stateList?: string[],
-	transitions?: Record<string, string[]>,
+	transitions?: Record<string, string | string[]>,
 ): StateComp {
 
 	if (!initState) {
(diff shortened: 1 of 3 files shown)


src/kaboom.ts
```diff
@@ -2159,16 +2163,34 @@ function state(initState: string, stateList?: string[]): StateComp {
...
+			const oldState = this.state;
+			checkTransition(oldState, state);
+			trigger("leave", oldState, ...args);
 			this.state = state;
-			trigger("enter", this.state, ...args);
+			trigger("enter", state, ...args);
+			trigger("enter", `${oldState} -> ${state}`, ...args);
+		},
+		onStateTransition(from: string, to: string, action: () => void) {
+			on("enter", `${from} -> ${to}`, action);
```
slmjkdbtl, line 2193: a bit hacky, but i kinda like it 🤔

src/kaboom.ts
```diff
@@ -2159,16 +2163,34 @@ function state(initState: string, stateList?: string[]): StateComp {
...
+	function checkTransition(from: string, to: string) {
+		if (!transitions?.[from]) {
+			return
+		}
+		const available = typeof transitions[from] === "string"
+			? [transitions[from]]
+			: transitions[from] as string[];
+		if (!available.includes(to)) {
+			throw new Error(`Cannot transition state from "${from}" to "${state}". Available transitions: ${available.map((s) => `"${s}"`).join(", ")}`);
+		}
```
masad-frost, line 2175: I wonder if it's more friendly if we return a boolean indicating success or failure.

src/kaboom.ts
masad-frost: Haha

src/kaboom.ts
```diff
@@ -2159,16 +2163,34 @@ function state(initState: string, stateList?: string[]): StateComp {
...
 	}
 
+	function checkTransition(from: string, to: string) {
+		if (!transitions?.[from]) {
+			return
+		}
+		const available = typeof transitions[from] === "string"
+			? [transitions[from]]
+			: transitions[from] as string[];
+		if (!available.includes(to)) {
```
masad-frost, line 2173: Seems unnecessary to create an array here.

```suggestion
		const isAvailable = typeof transitions[from] === "string"
			? transitions[from] === to
			: transitions[from].includes(to)
		if (!isAvailable) {
```

masad-frost (approved):
:pixel_heart:

src/kaboom.ts
slmjkdbtl: Absolutely

slmjkdbtl: Oh the array is used [later](https://github.com/replit/kaboom/blob/b0d1f50d68fb0f90d72bc325df5c480376bb222a/src/kaboom.ts#L2174) to print the available transition destinations

masad-frost: Gotcha

PR #428, opened 2021-11-16
[RFC] Automatically run onStateEnter() event for initial state

slmjkdbtl:
Before:

```js
state("idle", [ "idle", "attack", "move" ])

obj.onStateEnter("idle", () => {
    wait(1, () => obj.enterState("attack"))
})

// have to manually call enterState("idle") to trigger the onStateEnter() event defined above
obj.enterState("idle")
```

After:

```js
state("idle", [ "idle", "attack", "move" ])

// this will call the action immediately since our initial state is "idle"
obj.onStateEnter("idle", () => {
    wait(1, () => obj.enterState("attack"))
})
```

But consider this case:

```js
state("idle", [ "idle", "attack", "move" ])

obj.onStateEnter("attack", () => {
    wait(0.5, () => obj.enterState("move"))
})

// this will synchronously trigger the on enter "attack" we defined above, before executing on enter "idle" we defined below which should probably be triggered first for the initial state
obj.enterState("attack")

obj.onStateEnter("idle", () => {
    wait(1, () => obj.enterState("attack"))
})
```

### DIFF
--- a/demo/ai.js
+++ b/demo/ai.js
@@ -68,9 +68,6 @@ enemy.onStateUpdate("move", () => {
 	enemy.move(dir.scale(ENEMY_SPEED))
 })
 
-// Have to manually call enterState() to trigger the onStateEnter("move") event we defined above.
-enemy.enterState("move")
-
 // Taking a bullet makes us disappear
 player.onCollide("bullet", (bullet) => {
 	destroy(bullet)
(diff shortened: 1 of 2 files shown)


slmjkdbtl:
It's better to just do the enter action on next update

PR #429, opened 2021-11-16
Add outOfView comp

slmjkdbtl:
As discussed in #422 , would be nice to have a component for general out-of-view actions, including hiding, pausing, destroying, or custom event etc.

```js
const characterFarAway = add([
    pos(1200, 80),
    outOfView({ hide: true, pause: true }),
])
```

and `cleanup()` component can be just a thin wrapper of `outOfView({ destory: true })`

`outOfView` is slightly verbose comparing to other built-in comps, is there a more succinct name?


src/types.ts
```diff
@@ -3199,6 +3217,67 @@ export interface FollowComp extends Comp {
...
+	 */
+	pause?: boolean,
+	/**
+	 * If destroy object when out of view.
+	 */
+	destroy?: boolean,
+	/**
+	 * The screen bound offset.
+	 */
+	offset?: Vec2,
```
slmjkdbtl, line 3236: Not sure about this name, maybe call it `margin`?

lajbel:
suggestion: rename to out?

slmjkdbtl:
@lajbel An option, I wonder if people won't understand what it does from the name

lajbel:
> @lajbel An option, I wonder if people won't understand what it does from the name

outview()?

src/types.ts
lajbel: nop, offset is better (my oppinion)

slmjkdbtl:
> > @lajbel An option, I wonder if people won't understand what it does from the name
> 
> outview()?

yeah i kinda like that more than `outOfView`

lajbel (approved):
cool name

PR #431, opened 2021-11-17
[RFC] Add syntax to style chunks of text

slmjkdbtl:
Added a special syntax to enable style chunks of text
```js
drawText({
	text: "(oh hi):green this is a (styled):wavy text",
	styles: {
		"green": (idx, ch) => ({
			color: GREEN,
		}),
		"wavy": (idx, ch) => ({
			color: BLUE,
			pos: vec2(0, wave(-4, 4, time() * 4 + idx * 0.5)),
		}),
	},
})
```

Another idea is to use [BBCode](https://en.wikipedia.org/wiki/BBCode) syntax (implemented in #432)

```js
"[green]oh hi[/green] this is a [wavy]styled[/wavy] text"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - added `charSpacing` and `lineSpacing` in `TextCompOpt` and `DrawTextOpt`
 - added optional `transitions` argument in `state()` to define allowed transitions
 - added `StateComp#onStateTransition` to register event for specific transitions
+- added syntax for chunked text style `"this is a [styled].wavy text"` and `style` option in `TextCompOpt` and `DrawTextOpt` to define the styles with `CharTransformFunc`
 
 ### v2000.1.6
 
--- a/demo/text.js
+++ b/demo/text.js
@@ -89,7 +89,7 @@ onKeyDown("down", () => {
 
 // Use this syntax and style option to style chunks of text, with CharTransformFunc.
 add([
-	text("(oh hi):green this is a (styled):wavy text", {
+	text("[oh hi].green this is a [styled].wavy text", {
 		styles: {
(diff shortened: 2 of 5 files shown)


masad-frost (approved):
Didn't take long for you to create a templating engine. [Mail client](http://www.catb.org/jargon/html/Z/Zawinskis-Law.html) when?

I wonder if an approach like `chalk` is nicer, and also means we can have prebuilt styling.  Their tagged literals is similar to what you're going for too. I like how they allow you to apply multiple styles via dot separated values, in our case it'd look something like `"(oh hi):green.bold.wavy there`.

slmjkdbtl:
Ooh yeah it's quite similar to their tagged literal. Yes dot syntax for multiple is nice we should support that.

I like having option to prebuild too might save some performance on parsing for static strings. Maybe something like this?
```js
add([
    text(styled("(oh hi):green this is a (styled):wavy text"), { styles }),
])
```

Still looking if there's a better syntax tho. I guess the BBCode style won't make it if we have potentially long multi style tag. What do you think about this one

```js
"[oh hi].green.wavy this is a [styled].wavy text"
```

`[]` might be less usual in normal string compared to `()`, and use `.` instead of `:` for the first one

Mail client coming soon.

blurymind:
very nice! :) could we have it also work with xml style tags?

slmjkdbtl:
What kind of XML syntax are you thinking? I'm a bit hesitant to have a build-in XML parser. Maybe we can make it accept an intermediate format, like the type `compileStyledText()` returns, so people can build compiler for other syntax as third party?

blurymind:
no thats fine, you're right :) 

I can use some simple regex before feeding it to kaboom

PR #436, opened 2021-11-19
Rename dir() to vec2FromAngle()

slmjkdbtl:
`dir()` is unintuitive and takes a very generic name spot


masad-frost (approved):
:arrow_down:

PR #440, opened 2021-11-21
Merge multiple images to one sprite

tobibe:
Hello guys,

I recently wanted to create a character designer, where you choose your body in different colors and then choose beards, eyes, clothes, etc, like in https://sprite-generator.herokuapp.com/#

In Kaboom you can only load one image for any sprite, so you would have to create images for all possible characters combinations to be able to load them eventually.

So I extended loadSprite and loadSpriteAtlas to accept an array of url-strings and merge them together to one image, which is then treated as usual.

I created the demo (spriteMerged) where you can see two sprites (body and armor) and then the merged one and can move them.

Greetings


lajbel (approved):
Good!

slmjkdbtl (changes requested):
Hmm I'm not sure if `loadSpriteAtlas(string[])` is intuitive that it'll merge the images. 

`loadSpriteAtlas()` and `loadSprite()` can take `HTMLCanvasElement` as image source. What do you think make this a third party helper function like

```js
function mergeImg(imgs: Url[]): Promise<HTMLCanvasElement>
```

and do

```js
mergeImg(["body.png", "hat.png"]).then((canvas) => loadSpriteAtlas(canvas, cfg))
```

Also I think a more flexible way would be load them as separate images and draw them as separate sprites, so you can control their positions and colors etc dynamically (with #410 it should be easier to make a parent node with positions and add body and hat as children nodes). So I'm not sure to make image merging built-in as it's not the most advocated way to do this.

tobibe:
The mergeImg-function is definitly more intuitive than the array.

Should I remove the array-parameter from loadSprite and loadSpriteAtlas? Because it saves some code, if you know the trick.
And in both function, you can give only one sprite name, so it should be clear, that the result is one sprite. Only the way the function does it, is not obvious. 

I think the #410 is perfect for a swinging sword or shield and in this case, I would use two different loadSprite-calls. 
But if you have a hat, which is "walking" with the character, always stick to the same walking frame and changes it looks the same way as the characters frame, I think the rendering is faster, if you don't use the gameobject-tree and instead merge the images. 
And in the sword case, you maybe want a different colliding behavior, than the hat-case. So the objectgame-tree makes more sense.

slmjkdbtl:
You're right, if you don't need other things merging would be a better way. 

What I'm not sure about is if this should be core function bundled into kaboom, or should it be a third party helper 🤔

src/assets.ts
```diff
@@ -146,7 +147,7 @@ function loadAndMergeMultipleImgs(srcs: string[]): Promise<ImageData> {
 						ctx.drawImage(img, 0, 0);
 					}
                 });
-                resolve(ctx.getImageData(0, 0, images[0].width, images[0].height));
```
slmjkdbtl: oh sorry i missed this before, I think `ImageData` should be a better return type than `HTMLCanvasElement`

tobibe:
I removed the changes in loadSprite and type SpriteLoadSrc, exported the mergeImg-function and updated the example.


> What I'm not sure about is if this should be core function bundled into kaboom, or should it be a third party helper 🤔

Imho, if you don't include this function, the only "standard" option will be the gameobject-tree. Which will result in the example case (https://sprite-generator.herokuapp.com/#) in an object-tree with over 25 childs for character. If you construct not only your character, but randomized enemies this way, you easily get a big amout of components to handle it's animations and to render. (I know, that this will not be the standard case)
So hopefully, the developer thinks of a better ways to do it and checks in documentation, that the src-Attribute of loadSprite also accepts GfxTexData which itsself consists of ImageData. Now he has the option to search for an third party helper or write the merge-function by himself.

Long story short, I think, many people thinking about an alternative solution, will not find the fitting third party helper fast enough or think about the existance for this little specific function and stick to the tree, if they don't find a well documented function or an example, where they see this option. 

Maybe, if you provide a way to let these third party-helpers present themselves with examples near your kaboom documentation, like your "Playground".

slmjkdbtl:
I see, I thought it's more of a niche use case but maybe it's more common than I think. I think keep it in a demo would be a good for now 👍

tobibe:
Should I change something in my code?

slmjkdbtl:
Yeah, can you move the `mergeImg` function to `demo/spriteMerged.js`?

tobibe:
I have reverted my changed files and moved mergeImg to demo/spriteMerged.js

slmjkdbtl (approved):
Thank you! Sorry for the back and forth :D

PR #444, opened 2021-11-23
docs: Filter section names by search

mgmarlow:
The documentation navbar can be difficult to use on laptops or smaller
resolution devices due to needing to scroll down the list of section
names after a search query. Instead, filter the section names out of the
list if they don't have any matching entries.

## Before

image

## After

image

image


slmjkdbtl (approved):
Thank you! I like this.

Issue #450, opened 2021-11-28
onKeyRelease without parameters doesn't work


lajbel:
### Version

v2000.1

### What browsers are you seeing the problem on?

Chrome, Other

### What happened?

Basically, onKeyRelease without parameters does not run the function

### What's the expected behavior?

It is expected that when you release any key, the function will be executed.

### Minimum reproducable code

kaboom();

onKeyPress(() => console.log("pressed"));
onKeyRelease(() => console.log("released"));

### Live demo

https://kaboom-bug-1.lajbel.repl.co/

### Is there an existing issue for this?

- [X] I've searched and there is no existing issue for this bug

PR #453, opened 2021-12-01
fix: Allow onKeyReleased to check if any key is released

mgmarlow:
Fixes [#450](https://github.com/replit/kaboom/issues/450). Adds a flag, `isKeyReleased`, which enables listening to an event when any key is released, rather than only specific keys specified in the event handler.

This PR matches the implementation of `isKeyDown`.


slmjkdbtl (approved):
Nice! Thank you

PR #468, opened 2021-12-16
Disable automatic culling

slmjkdbtl:
Since #410 we're using shader uniform to apply camera instead of calculating the transformed vertices on CPU, the final vertices we calculate in `drawRaw` are world-space instead of screen-space due to this change, making the bound checking pointless if any camera transformation is involved. The direct effect is objects outside of [0, 0] - [width, height] were not rendered despite being in camera.

Should fix [this](https://github.com/replit/kaboom/pull/405#issuecomment-979701128)

### DIFF
--- a/src/gfx.ts
+++ b/src/gfx.ts
@@ -9,7 +9,6 @@ import {
 	isVec3,
 	isColor,
 	isMat4,
-	testRectRect,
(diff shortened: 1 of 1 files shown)


sergeichestakov (approved):
🧹

PR #491, opened 2022-01-23
Make esbuild not generate polyfills

slmjkdbtl:
The biggest source of kaboom slowness is esbuild generating polyfills for new js features like spread args, disabling it nearly doubles the performance. It's better to just don't use new features if there's breakage.


slmjkdbtl:
It's probably bad to do this for a minor release. Reverting and will apply for next major version

PR #498, opened 2022-01-26
render concept tutorials

sixhobbits:
this renders the concept tutorials in the site instead of linking out to github. 

it didn't seem to work if I put them under the /tutorials/ subfolder so /doc/ is a bit polluted now - hope that's ok.


slmjkdbtl:
`/doc` is a better place, thanks!

PR #511, opened 2022-01-31
Replace deprecated dir() with Vec2.fromAngle()

fabtjar:
I noticed it for https://kaboomjs.com/#vec2 and felt it should be up to date.


slmjkdbtl:
Thanks!

PR #556, opened 2022-03-07
New physics / collision detection system

slmjkdbtl:
## Changes

- Added support for `area()` and `body()` on nested children objects
- Added a new step `checkFrame()` between `updateFrame()` and `drawFrame()` which uses a spatial hash grid to check for collision between all objects, exposed a `hashGridSize` global option to control the hash grid size
- Supported areas for rotated rectangles (everything is a polygon now, however will support other shapes in the future)
- Switched to SAT for collision resolution
- Collision resolution is now performed automatically by `body()`, instead of only triggered by `PosComp#move()` or `Body#pushOut()`, added `isStatic` option to declare a body no affected by gravity and other objects (similar to `isStatic` in matterjs)
- Gravity is not enabled by default, giving `body()` to an object only makes it resolve collisions, call `gravity()` explicitly to enable gravity
- Added `Area#onCollisionEnd()` and `Area#onCollisionActive()` event
- Added `drawInspect()` event for components to draw whatever in inspect mode, moved area outlines to `AreaComp#drawInspect()`

## Known issues

- The new collision resolution doesn't prevent phasing through other objects at high speed (maybe use a sweeping check and pack that into `move()` like before, also like godot's `move_and_collide()`?)
- Since collision resolution (which might change object's `pos`) happens after all updates, stuff like `onUpdate(() => camPos(player.pos))` might be using the position before resolution, currently adding a `player.onCollisionResolve(() => camPos(player.pos))` to set cam pos to new position after resolution
- The current collision resolution doesn't do well with stacked objects (which is very common with `addLevel()`), results in some extra unwanted resolutions, here's an image I made to demonstrate the problem
col
- `Area#worldArea()` performs some matrix multiplications, it'll be nice to be cached
- Currently if 2 non-static bodies collide, the system will resolve the first object in the object pool, the behavior depends on the object order. What should be the desired behavior here? Previously it might be more "correct" because we're always resolving the moving object, because the collision check and resolution is triggered by `move()`, but now we don't have any info about movement, instead of maybe caching object position from last frame and check if changed. But again, the desired behavior should be more defined. Can resolve both object and use `mass` property to decide how much for each one, but the assumption is most people won't define a mass for ordinary cases.

## Open Questions
- Should the physics system manage a velocity on each game object? Can be handy for more realistic resolution? If so, should it be a field on `pos()` component that gets managed by `body()`?
- Introduce the concept of "force"? Like `applyForce()` in matterjs
- Allow custom collision resolver? Is there a way to let user to write their own physics, instead of making their own `body()`?
- Currently all collision events `onCollide()` `onCollisionActive()` `onCollisionEnd()` all happen when there's overlap, not when 2 objects are exactly on the edge of each other, should we add events for touching too, like `onTouch`? (e.g. those events would run after resolution, which will make 2 objects exactly on the edge of each other)
- `solid()` is removed in favor of `body({ isStatic: true })`, should we leave `solid()` as an alias for that instead of removing? Keeping it as alias and reducing one breaking change is nice, but `body()` is more explicit, let people know this is the single one component that deals with anything physicical
- This PR also removed `every(t)` and `revery(t)` which are aliases for `get(t).forEach()`, partly because of added `GameObj#getAll()` which is the recursive variant to `get()` and don't want to maintain more aliases like `everyAll()` and `reveryAll()`, also thought it can reduce the interface surface area and can be more intuitive to operate directly on the array from `get()`?
- Not 100% sure about some namings
  - `Area#onCollisionActive`, `Area#onCollisionEnd` (it has different prefix than `onCollide`, maybe they can be `onCollideActive` and `onCollideEnd`?)
  - `Area#collisionIgnore` for a list of tags to ignore when checking collision
  - `KaboomOpt#hashGridSize`
  - `GameObjRaw#getAll()` recursive version of `get()` to get all nested children objects, also removed `every()` in favor of `get(t).forEach` (or we'll need recursive versions like `everyAll()` and `reveryAll()`, thinking just removing them might be a cleaner interface?)


slmjkdbtl:
Going to merge this now because some other features in development depends on the new system and it makes things easier, going to address the issues gradually, as there's nothing very major now

PR #562, opened 2022-03-20
Add error message on object destruction failure

ma1ted:
Silent errors with things like this are just plain unhelpful

### DIFF
--- a/src/kaboom.ts
+++ b/src/kaboom.ts
@@ -3419,7 +3419,7 @@ export default (gopt: KaboomOpt = {}): KaboomCtx => {
 				if (this.parent) {
 					this.parent.remove(this)
 				} else {
-					console.error("Failed to destroy object")
+					throw new Error("Cannot to destroy parentless object.")
 				}
 			},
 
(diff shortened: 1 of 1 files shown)


src/kaboom.ts
```diff
@@ -3416,7 +3416,11 @@ export default (gopt: KaboomOpt = {}): KaboomCtx => {
 			},
 
 			destroy() {
-				this.parent?.remove(this)
+				if (this.parent) {
+					this.parent.remove(this)
+				} else {
+					console.error("Failed to destroy object")
```
slmjkdbtl, line 3422: Kaboom usually just throws error instead of only logging
```suggestion
					throw new Error("Cannot to destroy parentless object.")
```

slmjkdbtl (approved):
Yeah definitely should be a error message, just one small suggestion

slmjkdbtl:
Do you want to commit the suggestion and merge it?

ma1ted:
Apologies, I thought I already had. Done now.

PR #573, opened 2022-04-17
Completed `TODO: expose drawRaw`

a-a-GiTHuB-a-a:
I exposed `drawRaw` on the context object. (Is that enough?)


slmjkdbtl:
It's more of a design decision and have to ensure it's a good API for people to build upon. Still not entirely sure about its current form yet

PR #576, opened 2022-04-25
Use separate package for global type declarations

slmjkdbtl:
Currently kaboom always generate global type definitions, even when user is not using them by turning `global` option to `false`. Type defs for global variables should be optional too, or there'll be too much unused functions in global namespace

```js
const k = kaboom({
    global: false,
})

// tsc should catch this as error!
add()
```

This PR adds a separate package `kaboom-global` to contain the type defs for kaboom global functions, copies the code from `scripts/build.js` that generates global types to `pkgs/global/generate.js`

Currently this works by installing kaboom (the version of peer depended kaboom is synced with the version of `kaboom-global`) as peer dependency, then read `node_modules/kaboom/src/type.ts` to generate a `dist/kaboomGlobal.ts` which will be used as the typescript entry for the package

However this adds a little friction to users who want to use kaboom globals but also want types, that they have to `npm install kaboom-global`

### DIFF
--- a/pkgs/global/generate.js
+++ b/pkgs/global/generate.js
@@ -4,7 +4,7 @@ import ts from "typescript"
 
 const f = ts.createSourceFile(
 	"ts",
-	fs.readFileSync("../../src/types.ts", "utf-8"),
+	fs.readFileSync("node_modules/kaboom/src/types.ts", "utf-8"),
 	ts.ScriptTarget.Latest,
 	true,
 )
--- a/pkgs/global/generate.js
+++ b/pkgs/global/generate.js
@@ -1,0 +1,159 @@
+import fs from "fs"
+import path from "path"
+import ts from "typescript"
+
+const f = ts.createSourceFile(
+	"ts",
+	fs.readFileSync("../../src/types.ts", "utf-8"),
+	ts.ScriptTarget.Latest,
+	true,
+)
+
+function transform(o, f) {
+	for (const k in o) {
+		if (o[k] == null) {
+			continue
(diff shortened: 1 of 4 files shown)


sergeichestakov:
I'm confused - what's the downside of not shipping types with the package? Don't think we wanna cause unnecessary friction for users that do rely on them

masad-frost:
> I'm confused - what's the downside of shipping types with this package? Don't think we wanna cause unnecessary friction for users that do rely on them

Packages should not pollute global scope once you import them, unless they actually do modify global scope. In our case kaboom does not modify global scope unless you tell it to do so.  Right now, if you `import kaboom from 'kaboom'` all your modules will have functions like `add` despite you not instantiating kaboom, neither globally nor scoped.

masad-frost:
@slmjkdbtl what if we made it so that it just a sub path of kaboom package instead? `import 'kaboom/global'` it's an `.d.ts` file that imports defs from normal kaboom and puts it in global scope? 

People can do that easily without having to install another package. Also another added benefit is that you can reference `kaboom/global` from your tsconfig if you want it to be globally available in every module.

slmjkdbtl:
@sergeichestakov It's mainly for not polluting global name space when user is not using global functions, like in that code example it's nice to have `tsc` catch the undefined `add()` global function, however currently the global `add()` is always defined in kaboom.d.ts

@masad-frost  Oh didn't know you can do that, thanks I'll look into this way

slmjkdbtl:
#577

PR #577, opened 2022-04-26
Use subpath for global export

slmjkdbtl:
https://github.com/replit/kaboom/pull/576#issuecomment-1110123131

Use

```ts
import "kaboom/global"
```

to import defs for the global functions, instead of a separate package

Using typescript's [`typeVersions`](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#version-selection-with-typesversions) field to export types on a certain subpath (helped by the example [here](https://github.com/teppeis/typescript-subpath-exports-workaround))


masad-frost:
LGTM!

PR #601, opened 2022-06-19
support to pingpong option in `loadAseprite`

lajbel:
it's only one line, but this line could change everything...

* support to pingpong param in `loadAseprite`


DaInfLoop (approved):
I agree

slmjkdbtl:
Thanks! FYI pingpong isn't actually implemented yet tho

PR #602, opened 2022-06-19
doubleJump ghost comp removed

lajbel:
A strange component

* removed ghost comp `doubleJump`

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -52,7 +52,6 @@ export interface KaboomCtx {
 	 *     area(),
 	 *     body(),
 	 *     health(8),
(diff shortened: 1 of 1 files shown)


src/types.ts
```diff
@@ -52,7 +52,6 @@ export interface KaboomCtx {
 	 *     area(),
 	 *     body(),
 	 *     health(8),
-	 *     doubleJump(),
```
DaInfLoop, line 55: Could be a user-defined comp /shrug

lajbel: but it's confusing, since body() already has a doubleJump function

PR #605, opened 2022-06-23
Create SECURITY.md

Hugoonreplit:


lajbel:
I think this is not necessary because some bugs from previous versions are fixed in new versions. We do not maintain many versions at the same time.

PR #617, opened 2022-07-21
Fixed color lightening changing hue, as well as add screen color blending mode

seanlnge:
The method used to lighten and darken colors was implemented in a way that changed the hue, so I rewrote the method to lighten or darken colors by a percentage between 0-1, instead of adding a single number `n` to red, green, and blue.

I also implemented the [screen color blending mode](https://en.wikipedia.org/wiki/Blend_modes#Screen) using the method on wikipedia.


sravanthbaratam (commented):
LGTM!

PR #621, opened 2022-08-24
TODO: don't assume 1 as start opacity

lajbel:
Fixed todo :)

### DIFF
--- a/src/kaboom.ts
+++ b/src/kaboom.ts
@@ -4835,15 +4835,18 @@ export default (gopt: KaboomOpt = {}): KaboomCtx => {
 			throw new Error("lifespan() requires time")
 		}
 		let timer = 0
+		let startOpacity = 0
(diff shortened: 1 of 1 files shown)


lajbel:
Live demo: https://replit.com/@lajbel/Fix-opacity-thing?v=1

src/kaboom.ts
```diff
@@ -4835,15 +4835,18 @@ export default (gopt: KaboomOpt = {}): KaboomCtx => {
 			throw new Error("lifespan() requires time")
 		}
 		let timer = 0
+		let startOpacity = 0
```
slmjkdbtl, line 4838: I'd make this a `null` and check for `=== null` later, to say that the value doesn't exist yet instead of is number 0

slmjkdbtl:
Looks good! Thanks, just a small suggestion

PR #622, opened 2022-08-24
Update math.ts

mflerackers:
Added methods for cross product, angle between vectors and spherical linear interpolation (interpolation between directions)


mflerackers:
Cross product needs to be changed since it seems that Kaboom has y down, rotations clockwise

mflerackers (commented):
Fixed len/dist, added slen/sdist

slmjkdbtl:
Sorry missed this PR. Can you rebase and also update the type / doc in `types.ts`?

PR #623, opened 2022-08-26
added BodyComp#multiJump() to do multiple jumps easy

lajbel:
I think it's great that you can do multiple jumps in an easy way, but also if someone needs to do it manually, they can always use jump. 

Also added `Body#onMultiJump()` and `on("multiJump")` events.


slmjkdbtl:
I'm actually also thinking about separating `doubleJump` (and possibly multijump) to its own component like before, to teach more about components, wdyt?

lajbel:
would serve as an example, it would be great

lajbel:
closed in favor of `doubleJump` comp

PR #633, opened 2022-09-19
onHover fixed and format

lajbel:
yes


src/kaboom.ts
```diff
@@ -4023,6 +4011,21 @@ export default (gopt: KaboomOpt = {}): KaboomCtx => {
 						this.trigger("collideEnd", col.target, col)
 					}
 				}
+
+				if (this.isHovering()) {
```
slmjkdbtl, line 4015: I'll think about this more, `isHovering()` does intersection check and is a little expensive to add one call every frame to every object with area

lajbel: I see! then I will have to think of another solution

lajbel:
close in favor of #634

PR #634, opened 2022-09-20
Fix hover events

slmjkdbtl:
- Use closed variable in `Area#onHover()` and `Area#onHoverEnd()` instead of states
- Used another approach for global `onHover()` etc events to reduce all future custom logic for these global event handlers, see new helper `forAllCurrentAndFuture()`


lajbel:
cool 🐸

PR #638, opened 2022-09-26
Egg sprite

lajbel:
* egg
* egg
* egg

egg egg-crack


slmjkdbtl:
Nit: Let's use underscore `egg_crack.png` in file names instead of dashes `egg-crack.png`

PR #643, opened 2022-10-14
Added Scenes/SceneStay properties to stay() comp

nanopoison:
Scenes holds an array of scenes that will either stay in those scenes or get deleted in those scenes depending on the scene stay property.

### DIFF
--- a/src/kaboom.ts
+++ b/src/kaboom.ts
@@ -4962,10 +4962,12 @@ export default (gopt: KaboomOpt = {}): KaboomCtx => {
 		}
 	}
 
-	function stay(): StayComp {
+	function stay(scenes?: string[], sceneStay?: boolean): StayComp {
 		return {
 			id: "stay",
 			stay: true,
+			scenes: scenes,
+			sceneStay: sceneStay,
 		}
 	}
 
@@ -5185,9 +5187,21 @@ export default (gopt: KaboomOpt = {}): KaboomCtx => {
 
 			game.root.get().forEach((obj) => {
 				if (!obj.stay) {
-					game.root.remove(obj)
+					game.root.remove(obj);
+				} else {
+					var inScenes = false;
+					var s = obj.sceneStay != null ? obj.sceneStay : true;
+					if (obj.scenes != null) {
+						for (var scene of obj.scenes)
+							if (scene === id)
+								inScenes = true;
+						if ((!inScenes && s) || (inScenes && !s))
+							game.root.remove(obj);
+					} else {
+						game.root.remove(obj);
+					}
 				}
-			})
+			});
 
 			game.root.clearEvents()
 
--- a/src/kaboom.ts
+++ b/src/kaboom.ts
@@ -4962,12 +4962,11 @@ export default (gopt: KaboomOpt = {}): KaboomCtx => {
 		}
 	}
 
-	function stay(scenes?: string[], sceneStay?: boolean): StayComp {
+	function stay(scenes?: string[]): StayComp {
 		return {
 			id: "stay",
 			stay: true,
 			scenes: scenes,
-			sceneStay: sceneStay,
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


lajbel:
Very cool!

src/types.ts
```diff
@@ -4324,6 +4324,14 @@ export interface StayComp extends Comp {
 	 * If the obj should not be destroyed on scene switch.
 	 */
 	stay: boolean,
+	/**
+	 * Scenes that either destroy the object or stay (depending on sceneStay value)
+	 */
+	scenes: string[],
+	/**
+	 * Whether or not the object stays on a scene or gets destroyed on a scene
```
lajbel, line 4332: Maybe `sceneStay` isn't needed. Like if you add the scenes props, it's because you want only stay in that scenes, and if not, simple stay in all scenes :)

src/types.ts
```diff
@@ -4324,6 +4324,14 @@ export interface StayComp extends Comp {
...
+	 */
+	sceneStay: boolean,
 }
```
lajbel, line 4335: Maybe `sceneStay` isn't needed. Like if you add the scenes props, it's because you want only stay in that scenes, and if not, simple stay in all scenes :)

nanopoison: I thought it could be useful in the case that you wanted the **_ONLY_** delete the object if you are in those scenes instead of the other way around.

lajbel: some strange, I think it's more intuitive define the scenes for stay and not the scenes for not stay. What do you think @slmjkdbtl?

nanopoison: The default is for the scenes to stay, you have an option to make it not stay

lajbel: thinking, maybe `stay(["level1", "level2"])` and `destroy(["level1", "level2"])` or `leave(["level1", "level2"])`

Perhaps that would be more intuitive, `not stay` is like contradicting the name of the component itself

slmjkdbtl: I think passing in a `false` there is more confusing to users than writing out all the scenes to stay for in the first array

lajbel: Yes, fs `stay(scenes: string[])` is the best option to avoid confusions

nanopoison: So remove the `sceneStay` property, and add a `stayUntil(["level1"])` or similarly named component to have a similar effect while being more intuitive?

lajbel: nah, only `stay()` it's good for now :)

slmjkdbtl: Yeah I think one `stay(scenes: string[])` is great, thank you!!

lajbel (approved):

src/types.ts
```diff
@@ -4324,6 +4324,10 @@ export interface StayComp extends Comp {
 	 * If the obj should not be destroyed on scene switch.
 	 */
 	stay: boolean,
+	/**
+	 * Array of scenes that the obj will stay on.
+	 */
+	scenes: string[],
```
slmjkdbtl, line 4330: Can you rename this to something like `scenesToStay`? The property will be assigned to the game object directly, `obj.scenes` is a bit too ambiguous on what it does

nanopoison: Fair point, will do that real quick

src/kaboom.ts
```diff
@@ -5185,9 +5186,20 @@ export default (gopt: KaboomOpt = {}): KaboomCtx => {
...
+					if (obj.scenesToStay != null) {
+						var inScenes = false;
+						for (var scene of obj.scenesToStay)
+							if (scene === id)
+								inScenes = true;
+						if (!inScenes)
+							game.root.remove(obj);
+					} else {
+						game.root.remove(obj);
+					}
```
slmjkdbtl, line 5200: nit: I think these can be simplified to
```js
if (!obj.stay) {
	game.root.remove(obj)
} else {
	// if scenesToStay is defined, only stay on those scenes
	if (obj.scenesToStay != null && !obj.scenesToStay.contains(id)) {
		game.root.remove(obj)
	}
}
```
also can you run `npm run fmt`? Kaboom by default doesn't use semicolons and `fmt` can take care of that

src/types.ts
slmjkdbtl: Thank you!

src/kaboom.ts
nanopoison: Should be fixed & simplified. Hopefully isn't too simplified, could fix if needed

slmjkdbtl:
Good stuff, thank you! Merging now

slmjkdbtl:
Keep in mind if you add features in the future please also add to `CHANGELOG.md`, following the existing format!

nanopoison:
> Keep in mind if you add features in the future please also add to `CHANGELOG.md`, following the existing format!

Got it! Also... just realized that `scenes.js` was placed under `/src` lol...

slmjkdbtl:
Saw that too hh no worry removed that

PR #644, opened 2022-10-17
Headless mode and refactor

c-Gah:
Added initial support for headless mode. More refactoring will be needed, but this is a first step to running an authoritative server


lajbel:
aww, this is a BIG commit, thanks! But @slmjkdbtl should check it first! Thanks for the work anyway 💖💖💖

c-Gah:
```
import kaboom from 'kaboom';

// start the game
const boom = kaboom({ headless: true, width: 500, height: 500 })

// add character to screen, from a list of components
// In order to give the player a shape or size we need to create a rect for it
// Headless mode will not load sprites currently
const player = boom.add([
    //sprite("bean"),     // CURRENTLY IS BROKEN
    boom.rect(20, 20),
    boom.pos(120, 80),    // position in world
    boom.area(),          // has a collider
    boom.body(),          // responds to physics and gravity
])

boom.add([
    boom.rect(20, 20),
    boom.pos(160, 80),
    // Both objects must have area() component to enable collision detection between
    boom.area(),
    "enemy",
])

player.onUpdate(() => {
    player.move(10, 0);
});

player.onCollide("enemy", (enemy) => {
    console.log("Player collided with enemy")
})

player.onCollideEnd("enemy", (enemy) => {
    console.log("Player stoped colliding with enemy")
})
```

The above is an example of creating a game in headless mode. 
Functionally wise there is nothing else added. 
Headless mode also is functionally limited

c-Gah:
Also to note there is code redundancy and inefficiency present. The goal would be much smaller PRs in the future for further improvement, but this was about the smallest I could make this PR at this time

slmjkdbtl:
Hi! This is a big PR, can you explain a bit your idea behind this change? Which parts are added / removed / moved, what's the purpose of each file etc, thanks!

c-Gah:
Hello. Yes let me explain.

First to avoid breaking changes and too much at once I reverted the "front end" version of the code.

- `kaboom.ts`: The majority of this file in now unchanged. What was added was a headless option. When headless is flagged it returns a slimmed down version of kaboom. KaboomCoreCtx created and KaboomCtx extends it, hence the need for adding functions to the return object
- `classes.ts`: This file contains classes in `kaboom.ts`  ie. `Texture` and `SpriteData`
- `kaboomCore.ts`: This is mostly a slimmed down version of `kaboom.ts`
- `types.ts`: This is mostly separating out `KaboomCtx` and `KaboomCoreCtx ` as well as creating various Core types
- `utils.ts`: Just moving a function from `kaboom.ts` to `utils.ts` so that it can be referenced in `kaboom.ts` and `kaboomCore.ts`

PR #645, opened 2022-10-18
Part 1 - Refactor Kaboom classes

c-Gah:
This is a refactor for moving Classes outside of the main kaboom function

These classes will need to be referenced by functions outside of `kaboom.ts`

This is needed to eventually create a headless mode to be run server side. IE referenced from `kaboomCore.ts`

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -464,7 +464,7 @@ export interface KaboomCtx {
 	 *     player.hurt(1)
 	 *     bad.hurt(1)
 	 * })
-     *
+	 *
 	 * player.onCollide("apple", () => {
 	 *     player.heal(1)
 	 * })
@@ -2567,15 +2567,21 @@ export declare class SpriteData {
 	frames: Quad[]
 	anims: SpriteAnims
 	constructor(tex: Texture, frames?: Quad[], anims?: SpriteAnims)
-	static fromImage(data: TexImageSource, opt?: LoadSpriteOpt): SpriteData
-	static fromURL(url: string, opt?: LoadSpriteOpt): Promise<SpriteData>
+	static fromImage(data: TexImageSource, gl: WebGLRenderingContext, gc: (() => void)[], gopt: KaboomOpt, slice: any, opt?: LoadSpriteOpt): SpriteData
+	static fromURL(url: string, loadImg, gl: WebGLRenderingContext, gc: (() => void)[], gopt: KaboomOpt, slice, opt?: LoadSpriteOpt): Promise<SpriteData>
+}
+
+export type AudioData = {
+    ctx: AudioContext;
+    masterNode: GainNode;
+    burpSnd: SoundData;
 }
 
 export declare class SoundData {
 	buf: AudioBuffer
 	constructor(buf: AudioBuffer)
-	static fromArrayBuffer(buf: ArrayBuffer): Promise<SoundData>
-	static fromURL(url: string): Promise<SoundData>
+	static fromArrayBuffer(buf: ArrayBuffer, audio: AudioData): Promise<SoundData>
+	static fromURL(url: string, audio: AudioData, fetchArrayBuffer): Promise<SoundData>
 }
 
 export interface LoadBitmapFontOpt {
@@ -2704,11 +2710,14 @@ export type TextureOpt = {
 }
 
 export declare class Texture {
+	gl: WebGLRenderingContext
+	gc: (() => void)[]
+	gopt: KaboomOpt
 	glTex: WebGLTexture
 	width: number
 	height: number
-	constructor(w: number, h: number, opt?: TextureOpt)
-	static fromImage(img: TexImageSource, opt?: TextureOpt): Texture
+	constructor(w: number, h: number, gl: WebGLRenderingContext, gc: (() => void)[], gopt: KaboomOpt, opt?: TextureOpt)
+	static fromImage(img: TexImageSource, gl: WebGLRenderingContext, gc: (() => void)[], gopt: KaboomOpt, opt?: TextureOpt): Texture
 	update(x: number, y: number, img: TexImageSource)
 	bind()
 	unbind()
--- a/src/classes/Asset.ts
+++ b/src/classes/Asset.ts
@@ -1,0 +1,53 @@
+import { Event } from "../utils"
+
+export class Asset<D> {
+	done: boolean = false
+	data: D | null = null
+	error: Error | null = null
+	private onLoadEvents: Event<[D]> = new Event()
+	private onErrorEvents: Event<[Error]> = new Event()
+	private onFinishEvents: Event<[]> = new Event()
+	constructor(loader: Promise<D>) {
+		loader.then((data) => {
+			this.data = data
+			this.onLoadEvents.trigger(data)
+		}).catch((err) => {
+			this.error = err
+			if (this.onErrorEvents.numListeners() > 0) {
+				this.onErrorEvents.trigger(err)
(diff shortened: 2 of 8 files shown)


src/types.ts
```diff
@@ -2704,11 +2710,14 @@ export type TextureOpt = {
...
+	gl: WebGLRenderingContext
+	gc: (() => void)[]
+	gopt: KaboomOpt
 	glTex: WebGLTexture
 	width: number
 	height: number
-	constructor(w: number, h: number, opt?: TextureOpt)
-	static fromImage(img: TexImageSource, opt?: TextureOpt): Texture
+	constructor(w: number, h: number, gl: WebGLRenderingContext, gc: (() => void)[], gopt: KaboomOpt, opt?: TextureOpt)
+	static fromImage(img: TexImageSource, gl: WebGLRenderingContext, gc: (() => void)[], gopt: KaboomOpt, opt?: TextureOpt): Texture
```
slmjkdbtl, line 2720: If we're exporting `Texture` I think we shouldn't need user to pass `gl`, `gc` and `gopt`, we can just not export `Texture` in non-browser environment since it wouldn't make sense anyway

c-Gah: When a user adds a Sprite with this command `loadSprite("bean", "/sprites/bean.png")`
The function call will execute `SpriteData.fromURL(src, loadImg, gl, gc, gopt, slice, opt)` where `gl`, `gc`, `gopt` are all already defined above. Users will not have to pass `gl`, `gc`, `gopt`

It's true most `Texture` could just be not exported for non-browser environments. Non-browser environments will probably use none of the `Asset` `SpriteData` or `Texture` classes

However I think Kaboom is suffering from a God Object problem:
https://en.wikipedia.org/wiki/God_object

Because most of the functionality is in `kaboom.ts`. It makes changes like adding a headless mode take a long time. 

Ideally `kaboom.ts` could be condensed from ~6,600 lines of code to ~1,000 lines of code

There is also another small issue in `kaboom.ts` I've discovered while reading through it. Some variables are declared out of order, but it's not as apparent because hoisting fixes it. This can lead to more unintended consequences 
https://www.w3schools.com/js/js_hoisting.asp

slmjkdbtl: Yes it wouldn't affect calling `loadSprite`, but it's planned to export the `Texture` class to have finer control over textures especially when 3D comes out. Part of the reason Kaboom switched to this one large file format is to make sharing closed variables easier, so don't have to pass a bunch of state objects around. It makes it harder to navigate one large file but I think we shouldn't change the interfaces when trying to split it.

My idea of adding headless mode was to identify all the browser dependencies, like all uses of `window` `document` `navigator` etc and abstract those into an `Engine` interface put them into a `browser.ts`, implement that same interface in a `node.ts` (there'll probably be a lot of noop), so `kaboom.ts` and all other files won't have any browser dependencies and be runnable in any js environment, and import `browser.ts` or `node.ts` based on the current environment. This will also help us move to other js platforms than browser or node, if there are any. Wdyt?

c-Gah: See: https://blog.sessionstack.com/how-javascript-works-the-factory-design-pattern-4-use-cases-7b9f0d22151d

"**Factory pattern**: we create objects without exposing the creation logic to the client and the client uses the same common interface to create a new type of object."

I think technically you are employing the Factory Design pattern right now. `Texture` executes the line `this.glTex = gl.createTexture()` on the first line in the `constructor`

The `Texture` class needs `gl` in order to create a texture, but you are not exposing it to the client because `kaboom()` handles it for the client. If we separated out the code into something simple like below the logic would be identical, while also allowing the bulk of `Texture` class code to be in a separate file:

```
function Texture(width, height): TextureClass {
    return new TextureClass(width, height, gl, gc, gopt)
}

// the exported ctx handle
const ctx: KaboomCtx = {
    Texture,
```
or even in the separate `Texture` file you could do a
```
import { gl } from './kaboom'

export class Texture {
    glTex: WebGLTexture
    constructor(w: number, h: number, opt: TextureOpt = {}) {
        this.glTex = gl.createTexture()
```

c-Gah: Regarding abstracting into `Engine` and switching between `browser.ts` and `node.ts` I believe that would work. However it seems like a lot of work

I also imagine doing a `noop` on all of `window` `document` `navigator` would result in a lot more code

I think doing something like below would result in less code and you wouldn't need to `noop` `canvasEvents` etc. Functionality could switch by reading a headless flag in `KaboomOpt `
```
class KaboomCore {
    gl: WebGLRenderingContext
    gc: (() => void)[]
    gopt: KaboomOpt
```

```
class Kaboom extends KaboomCore {
    canvasEvents: EventList<HTMLElementEventMap>
    docEvents: EventList<DocumentEventMap>
    winEvents: EventList<WindowEventMap>
````
you could also keep it functional and extending kaboomCore by doing
```
export default (gopt: KaboomOpt = {}): KaboomCoreCtx  | KaboomCtx => {
	const { gl, gc } = { ...kaboomCore(gopt) }
```

Comment (slmjkdbtl) [reply src/types.ts:2702]: Do you think you can make a runnable demo with your approach, like a basic POC multiplayer game, including the final UX of using headless kaboom? For big changes like this it's hard to make decisions without seeing how it'll be used in the end

Comment (c-Gah) [reply src/types.ts:2702]: To specify above you would implement a `kaboomManager`

```
if (gopt.headless) {
    return kaboomCore(gopt)
else 
    return kaboom(gopt)
}
```

Node won't error out because it doesn't read the `window` and `document` variables in `kaboom.ts`

Comment (c-Gah) [reply src/types.ts:2702]: It is definitely a big change. I am still working on refactoring headless. But my version 1 will be something like below

import kaboom from 'kaboom';

`server.ts`
```
// start the game
const boom = kaboom({ headless: true, width: 500, height: 500 })

boom.initNetwork()

const testPlayer = boom.add([
    boom.rect(20, 20),
    boom.pos(120, 80),    // position in world
    boom.area(),          // has a collider
    boom.body(),          // responds to physics and gravity
])

var players = [ testPlayer ]

players.forEach((player) = {
    // Move all the players to laugh at them
    player.onUpdate(() => {
        player.move(10, 0);
    });
});

boom.onUpdate(() => {
    // Updates all connected clients with new player positions
    // This would be configured to fire maybe 15 times a second by default
    boom.updateClients({ "players": players })
});
```

`client.ts`
```
kaboom()

initNetwork("localhost:5000")

loadSprite("bean", "/sprites/bean.png")

const player = add([
    sprite("bean"),
    // center() returns the center point vec2(width() / 2, height() / 2)
    pos(center()),
])

onKeyDown("left", () => {
    //This will call the move method as normal as well as send data through the open network channel
    player.networkMove(-SPEED, 0)
})

// Create the logic here to update all players positions upon receiving server updates
onNetworkUpdate(() => {
});
```

c-Gah: The syntax and naming may be a bit different, but that's the general idea

Version 2 would be building out more basic and advanced network features such as Snapshot Interpolation:
https://github.com/geckosio/snapshot-interpolation#readme

c-Gah: My vision is that Kaboom certainly isn't as powerful or feature complete as some other JavaScript game engine. But few have as clean of syntax as kaboom and even fewer have server and network functionality in mind. Kaboom could fit a niche market for those who want a low code, simplistic engine with integrated networking, and other features not provided with other engines

slmjkdbtl: Thanks for the code demo! Do you also have a runnable demo on your branch?

PR #650, opened 2022-11-04
adds shift, control, and alt to Key type

iamthe-Wraith:
The keys: 'shift', 'control', and 'alt' were missing from the Key type which was causing errors in typescript.


lajbel:
Awesome! Thannks :D

lajbel (approved):

PR #652, opened 2022-11-21
Fixed mousePos() and onClick() to also work on Firefox

Nassim-Saboundji:
In kaboom@3000.0.0-alpha.14 `mousePos()` and by extension `onClick()` where broken on Firefox. The reason turned out that `e.offsetX` and `e.offsetY` on firefox are always 0. This PR swaps the use of `e.offsetX` and `e.offsetY` with `e.x` and `e.y`. 

This fix works both on Firefox while maintaning functionality on chromium based browsers.

Created a test in `onClick.js` to make sure everything works all right.

### DIFF
--- a/examples/onClick.js
+++ b/examples/onClick.js
@@ -1,0 +1,12 @@
+kaboom()
+
+loadSprite("bean", "/sprites/bean.png")
+const player = add([sprite("bean"), area(), pos(center()), "player"])
+
+player.onClick(() => {
+  debug.log("clicked!")
+})
+
+onClick("player", () => {
+  debug.log("clicked!")
+})
--- a/src/kaboom.ts
+++ b/src/kaboom.ts
@@ -2968,7 +2968,7 @@ export default (gopt: KaboomOpt = {}): KaboomCtx => {
 
 	canvasEvents.mousemove = (e) => {
 		game.ev.onOnce("input", () => {
-			setMousePos(e.offsetX, e.offsetY)
+			setMousePos(e.x, e.y)
 			game.ev.trigger("mouseMove")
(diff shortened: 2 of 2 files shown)


Nassim-Saboundji:
I don't know what is happening with eslint?

slmjkdbtl:
Thanks Nassim! Also noticed this problem

I think `x` (alias to `clientX`) is not what we want here here, it gives the coordinate relative to viewport instead of the canvas element so the mouse pos won't be correct if the canvas isn't at the top left corner of the viewport.

I think a quick fix for now could be

```js
canvasEvents.mousemove = (e) => {
	const [x, y] = [e.offsetX, e.offsetY]
	game.ev.onOnce("input", () => {
		setMousePos(x, y)
		game.ev.trigger("mouseMove")
	})
}
```

the bug is it looks like firefox modified `e.offsetX` to `0` before that `"input"` event is run, not sure if it's a firefox update or some recent changes caused this going to look into more

slmjkdbtl (changes requested):
See comment above

slmjkdbtl:
ESLint is wanting you to change spaces to tabs (kaboom uses tabs here), you can do a quick `npm run fmt` before commit!

slmjkdbtl:
I think the break started in aa0aacf5e2e082ef03c909930de0e300c70b9182 where we delayed the input handling to the "input" phase each frame and wasn't aware firefox will change the event object after the event is dispatched.

Anyway had to do this fix myself to publish a quick fix e7ab74c5af8eaf9ec69dc5ebdc017ef95c76657e, thanks for helping on the issue!

PR #658, opened 2022-12-14
Updated math with more methods

mflerackers:
Reflect, project and reject vectors. Spherical linear interpolation for direction vectors. Cross product and angle between two direction vectors.


slmjkdbtl:
These look great, thank you!

PR #660, opened 2022-12-18
Pathfinding

mflerackers:


lajbel:
Kaboom doesn't use `;`, you should run `npm run lint`. Epic commit!

mflerackers:
Yeah, I was thinking that. I also need to know what to do with private interfaces.

src/kaboom.ts
```diff
@@ -5467,11 +5477,335 @@ export default (gopt: KaboomOpt = {}): KaboomCtx => {
...
 		return {
 			id: "obstacle",
 			require: [ "tile" ],
-			isObstacle: true,
+			setObstacle(this: GameObj<ObstacleComp | TileComp>, isObstacle: boolean) {
+				if (_isObstacle !== isObstacle) {
+					_isObstacle = isObstacle
+					this.getLevel().invalidateNavigationMap()
+				}
+			},
```
slmjkdbtl, line 5490: Is it possible to use getter / setter here?

src/kaboom.ts
```diff
@@ -5467,11 +5477,335 @@ export default (gopt: KaboomOpt = {}): KaboomCtx => {
...
+				const goal = this._tile2Hash(to)
+	
+				// Tiles are not accessible
+				// If we test the start tile, we may get stuck
+				/*if (_costMap[start] === Infinity) {
+					return null
+				}*/
+				if (_costMap[goal] === Infinity) {
+					return null
+				}
```
slmjkdbtl, line 5582: What about an an option to navigate to the closest non-obstacle tile, is that common?

mflerackers: Changed it. Didn't know we used get/set from reading the Kaboom2000 code. I like get/set better for most things.

src/kaboom.ts
mflerackers: Changed it. Didn't know we used get/set from reading the Kaboom2000 code. I like get/set better for most things.

src/kaboom.ts
mflerackers: I'll look into it. With the raycasting component you can look for the closest accessible (from your position) tile on a straight line towards the goal. But A* pathfinding will just hang if you feed it inaccessible tiles, therefore you check the connectivity map first. You could use breadth first search and then choose one of the closest tiles, but it might be slow.

slmjkdbtl: Definitely not a required feature for first launch, don't have a good idea on how to choose the closest point now either. Can also pick one from the neighbors of the goal tile, that's also has the closest distance to the start tile?

src/kaboom.ts
```diff
@@ -5467,11 +5477,335 @@ export default (gopt: KaboomOpt = {}): KaboomCtx => {
...
+						n.push(node + 1)
+					}
+					if (bottom) {
+						n.push(node + numColumns)
+					}
+				}
+				return n
+			},
+	
+			_getCost(this: GameObj<LevelComp | PathfindingComp>, node: number, neighbour: number) {
```
slmjkdbtl, line 5704: Is `node` parameter unused here?

src/kaboom.ts
```diff
@@ -5467,11 +5477,335 @@ export default (gopt: KaboomOpt = {}): KaboomCtx => {
...
+			id: "edges",
+			require: ["tile"],
+			mask: _mask,
+		}
+	}
+
+	function pathfinding() : PathfindingComp {
+		let _costMap: number[] | undefined
+		let _edgeMap: number[] | undefined
+		let _connectivityMap: number[] | undefined
```
slmjkdbtl, line 5527: Nit: Can we type them to `number[] | null` here and initialize to `null`?

src/kaboom.ts
```diff
@@ -5467,11 +5477,335 @@ export default (gopt: KaboomOpt = {}): KaboomCtx => {
...
+							const object = objects[j]
+							if (object.isObstacle !== undefined) {
+								if (object.isObstacle) {
+									cost = Infinity
+									break
+								}
+							}
+							else {
+								cost += object.cost || 0
+							}
```
slmjkdbtl, line 5744: Is it ok to simplify it to
```js
cost += object.isObstacle ? Infinity : (object.cost ?? 0)
```

src/kaboom.ts
```diff
@@ -5447,6 +5453,10 @@ export default (gopt: KaboomOpt = {}): KaboomCtx => {
 				)
 			},
 
+			getLevel() {
+				return level
+			},
+
 			moveLeft() {
 				this.setTilePos(this.tilePos.add(vec2(-1, 0)))
 			},
```
slmjkdbtl, line 5462: Think we can integrate obstacles into these `move` functions

src/kaboom.ts
mflerackers: Currently yes, because we only take the destination cost into account. In some graphs you may take both source and destination into account. For example if you have a height map, 1 to 6 means a slope of 5 upwards. In other cases the cost may be stored in edges of the graph rather than the nodes, so you might need both nodes to index the edge.

src/kaboom.ts
mflerackers: Sure

src/types.ts
```diff
@@ -4533,19 +4557,65 @@ export interface LevelOpt {
...
+
+export interface EdgesComp extends Comp {
+    mask: EdgeMask
+}
+
+export interface PathfindingComp extends Comp {
+	invalidateNavigationMap(),
+	onNavigationMapChanged(cb: () => void): EventController,
+	getTilePath(from: Vec2, to: Vec2, diagonals?: boolean): Vec2[],
+	getPath(from: Vec2, to: Vec2, diagonals?: boolean): Vec2[],
```
slmjkdbtl, line 4612: Nit: these return `Vec2[] | null`

src/types.ts
```diff
@@ -4533,19 +4557,65 @@ export interface LevelOpt {
...
+    RightVertical = 14,
+    All = 15,
+}
+
+export interface EdgesComp extends Comp {
+    mask: EdgeMask
+}
+
+export interface PathfindingComp extends Comp {
+	invalidateNavigationMap(),
```
slmjkdbtl, line 4609: Should users call this manually?

src/types.ts
```diff
@@ -4533,19 +4557,65 @@ export interface LevelOpt {
...
+	onNavigationMapChanged(cb: () => void): EventController,
```
slmjkdbtl, line 4610: When would a end user use `invalidateNavigationMap()` manually, or listen to `onNavigationMapChanged()`?

src/kaboom.ts
mflerackers: You would need at least

cost += object.isObstacle !== undefined ? (object.isObstacle ? Infinity : 0) : (object.cost ?? 0)
And you'd lose the early exit

src/types.ts
mflerackers: That's true

src/types.ts
mflerackers: Agent listens to it, because it needs to know when its path became invalid and needs to be recomputed. A user building a bot might need to know it too.
invalidateNavigationMap is not private since theoretically you can make changes to the level which can't be detected by us.

src/kaboom.ts
```diff
@@ -5563,9 +5908,110 @@ export default (gopt: KaboomOpt = {}): KaboomCtx => {
...
+						spatialMap[hash].splice(index, 1)
+					}
+				}
+			},
 		}
 
 		level.use(levelComp)
+		if (opt.usePathfinding) {
+			level.use(pathfinding())
+		}
```
slmjkdbtl, line 6014: I'm thinking if we can omit this option, and use pathfinding if user call any of the pathfinding functions, so from the UX side user have one less option to care about and can always assume addLevel has pathfinding capabilities

src/types.ts
slmjkdbtl: What kind of changes we can't detect?

src/kaboom.ts
slmjkdbtl: You meant early exit for `object.isObstacle !== undefined`? Yeah cuz I was also wondering if `isObstacle = false` tile can also have a `cost`

mflerackers: A user could use both cost and obstacle yes

if (object.isObstacle !== undefined && object.isObstacle) {
    cost = Infinity;
    break;
}
else {
    cost += object.cost || 0;
}

slmjkdbtl: Oh you meant the early `break` sorry

src/types.ts
mflerackers: A user can remove the obstacle component for example. Or Add an isObstacle: true without using obstacle(). Or even a get isObstacle() in case the object determines the obstacle state from other states.

src/kaboom.ts
mflerackers: We could test this in getPath/getTilePath and add the component if needed yes.

src/kaboom.ts
mflerackers: Yes, once you find an obstacle, you don't need to check other objects anymore.

slmjkdbtl (approved):
Looks good besides some nitpicks, I think we can merge now but thinking to add these soon 
later after this PR:
- I think we can merge `tile`, `cost`, `obstacle` and `agent`
- Should support obstacle checking in `moveLeft()`, `moveRight()` etc
- Omit `usePathfinding` option and initiate stuff when user call any of the pathfinding functions

Let me know if you finished and it's good to merge

src/kaboom.ts
mflerackers: I think we can use the connectivity map to look for the point in the current region which is closest to the target. However, that point may not always make sense. If you have a wall with a closed door, and the door has the possibility to connect the current region with a region which is closer to the target, it makes more sense to walk and stop in front of the door than towards a wall which won't budge.
So you would need to know potential connections somehow, obstacles which have the potential to lose their obstacle status and let you pass.
But if you know which obstacles can be unblocked, the best strategy would be to calculate the path as if those obstacles don't exist. This way the agent won't walk towards the door of a room which has no other exits, which would happen if it just looked for the closest position.

```
|---------|
|  x      |
|----|--d-|
|    |    |
|--d-|--d-|
| a       |
|------d--|
```

PR #671, opened 2023-01-27
Fixed testLineCircle

mflerackers:


sravanthbaratam (commented):
@mflerackers Can you tag the issue that is being resolved by this PR?

PR #698, opened 2023-03-31
Kaboom gamepads

lajbel:
There's the update of the first gamepad pr.

Now Kaboom has a `mergedGamepadState`, where is combined all gamepads inputs, and a `gamepadsState`, where is each gamepad state. Now, `getGamepads()` works, and returns every gamepad registered by Kaboom.

```js
const gamepads = getGamepads();

gamepads[0].isPressed(); 
gamepads[1].isDown();
gamepads[2].isReleased();
gamepads[3].getStick();
gamepads[4].index // the index in the browser count of gamepads
```

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -48,6 +48,11 @@ export class ButtonState<T = string> {
 	}
 }
 
+class GamepadState {
+	buttonState: ButtonState<GamepadButton> = new ButtonState()
+	stickState: Map<GamepadStick, Vec2> = new Map()
+}
+
 class FPSCounter {
 	private dts: number[] = []
 	private timer: number = 0
@@ -90,7 +95,9 @@ export default (opt: {
 		mouseDeltaPos: new Vec2(0),
 		keyState: new ButtonState<Key>(),
 		mouseState: new ButtonState<MouseButton>(),
-		gamepadButtonState: new ButtonState<GamepadButton>(),
+		mergedGamepadState: new GamepadState(),
+		gamepadStates: new Map<string, GamepadState>(),
+		gamepads: [] as KGamePad[],
 		charInputted: [],
 		isMouseMoved: false,
 		lastWidth: opt.canvas.offsetWidth,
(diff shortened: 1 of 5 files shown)


lajbel:
Now, the `KaboomCtx` methods like `onGamepadXXXX`, get info from the merged gamepad state, but should it have an option to select a gamepad? Ex `onGamepadButtonDown("dpad-left", 1)`, for controller 2.

Also, I have this idea for some functions

```js
const playerOneGamepad = getGamepads()[0];

playerOneGamepad.isStickDown(); // and for every side
playerOneGamepad.isSideDown(); // equal to .isButtonDown("dpad-down") && .isButtonDown("south")
```

src/app.ts
```diff
@@ -90,7 +95,9 @@ export default (opt: {
 		mouseDeltaPos: new Vec2(0),
 		keyState: new ButtonState<Key>(),
 		mouseState: new ButtonState<MouseButton>(),
-		gamepadButtonState: new ButtonState<GamepadButton>(),
+		mergedGamepadState: new GamepadState(),
```
slmjkdbtl, line 98: I think it's more intuitive to have the global gamepad apis default to gamepads[0] instead of a merged state, wdyt?

lajbel: Hmm I think gamepads[0] is more for detect the real gamepads of the user, I imagine the user's side trying to get the first gamepad using `[0]` and the second using `[1]`, and see how it's possible trigger both with the second controller. But if you want, I can change it, obviously.

slmjkdbtl: What do you mean by get the real gamepads of the user and trigger both with the second controller?

I was imagining `isGamepadPressed()` to be a convenience function for `getGamepads()[0].isPressed()` instead of checking "if any gamepad has a button pressed", mainly because I can't think of a case where a game actually want to use a merged gamepad state when multiple gamepads are connected. Not sure, maybe let's merge this for now and decide later.

lajbel: Hmm, yes, maybe decide it later, I think there can be some persons who will want to read the input of all gamepads, for example for one player games. Anyway, both sounds good for me, use `isGamepadXXXX` and `onGamepadXXXX` as alias for `gamepads[0].isXXXX`.

PR #703, opened 2023-04-12
Fix file type in setup.md.

MDxWARRIORxOP:
The file type of the horse sound was wrong (png), so i changed it to mp3.

also fixed a typo.


slmjkdbtl:
thx!

PR #730, opened 2023-06-18
empty components and plugins with options

### LINKED ISSUE #691: TS `move()` comp returns `Comp` type, and that "bugs" `add()` return type
The `move` component typescript returns a `Comp` type

image

And that makes the type returned by `add()` be only `Game<Comp>`

image

If I remove `move()` component, this is transformed in 

image

I think the fix is to make an extended `Comp` type for `move()`, such `MoveComp`.

lajbel:
### No plugins, no type
If there is not a plugins property inside `KaboomOpt`, it only returns `KaboomCtx`

**Before**

image
**Now**

image

### Options in plugins
Now there's a support for options in plugins

**Normal plugins**
```ts
function logPlugin(k: KaboomCtx) {
    return {
        logWidth() {
            return {
                id: "logWidth",
                log() {
                    k.debug.log(k.width());
                },
            };
        },
        logHeight() {
            return {
                id: "logHeight",
                log() {
                    k.debug.log(k.height());
                },
            };
        },
    };
}
```

**Plugins with options**
```ts
function newLogPlugin(log: "width" | "height") {
    return (k: KaboomCtx) => {
        return {
            logRes() {
                return {
                    id: "hello",
                    log() {
                        k.debug.log(k[log]());
                    },
                }
            },
        };
    };
}
```

They work both
image

### Empty Component type

This was a big odyssey, basically I discovered `MergeObj<T>` doesn't work good when there are types with not required properties (for example, `Comp` type), so my first solution was doing this

```ts
type LifespanComp = Comp & { id: string }
type MoveComp = Comp & { id: string }
```

This fixes the problem without don't have required properties, because now id is required, and we got this result

image

But if there's two duplicated values in an Union, typescript ignores one, so this can happen

image

One is ignored, and looks strange, so I made `EmptyComp`, it represents a Comp with an id, but without nothing more, so, if there are 2 components that returns an `EmptyComp`, it makes sense with the returned object type. It should fix #691

image

### DIFF
--- a/src/kaboom.ts
+++ b/src/kaboom.ts
@@ -112,7 +112,6 @@ import type {
 	AnchorComp,
 	ZComp,
 	FollowComp,
-	MoveComp,
 	OffScreenCompOpt,
 	OffScreenComp,
 	AreaCompOpt,
@@ -136,12 +135,12 @@ import type {
 	FixedComp,
 	StayComp,
 	HealthComp,
-	LifespanComp,
 	LifespanCompOpt,
 	StateComp,
 	Debug,
 	KaboomPlugin,
 	MergeObj,
+	EmptyComp,
 	LevelComp,
 	Edge,
 	TileComp,
@@ -3734,7 +3733,7 @@ export default (gopt: KaboomOpt = {}): KaboomCtx => {
 		}
 	}
 
-	function move(dir: number | Vec2, speed: number): MoveComp {
+	function move(dir: number | Vec2, speed: number): EmptyComp {
 		const d = typeof dir === "number" ? Vec2.fromAngle(dir) : dir.unit()
 		return {
 			id: "move",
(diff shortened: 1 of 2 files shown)


src/kaboom.ts
```diff
@@ -477,7 +476,9 @@ export default (gopt: KaboomOpt = {}): KaboomCtx => {
 			const tex = new Texture(0, 0, opt)
 			tex.bind()
 			gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE, img)
+			// @ts-ignore
 			tex.width = img.width
+			// @ts-ignore
```
slmjkdbtl, line 481: why do we have to ignore here?

slmjkdbtl: tried remove them and don't get any warnings

lajbel: I see this one image

slmjkdbtl:
Thank you this is amazing

PR #732, opened 2023-06-21
Expose clamp() math helper

luizbills:


slmjkdbtl:
I agree to expose it, but we should also add this function and the type sig to the types (`src/types.ts`) so it's in LSP and docs, you can add it [here](https://github.com/replit/kaboom/blob/2cb340f7d06d53fb98a7ca90105f442746739fa1/src/types.ts#L1843)

luizbills:
> I agree to expose it, but we should also add this function and the type sig to the types (`src/types.ts`) so it's in LSP and docs, you can add it [here](https://github.com/replit/kaboom/blob/2cb340f7d06d53fb98a7ca90105f442746739fa1/src/types.ts#L1843)

Done.

slmjkdbtl:
amazing, thx

PR #743, opened 2023-07-16
Update types.ts - documentation comment for Texture.free()

codingMASTER398:
Adds a comment explaining the usage of Texture.free() and it's importance


slmjkdbtl:
Thanks! Yes it should be noted to be called manually.

PR #771, opened 2023-10-16
Fix emoji display in text

luluxia:
Native `String split()` can't handle emoji in strings correctly so it causes emoji display exception in text component.
I added a library to fix this.

image

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -297,3 +297,203 @@ export class BinaryHeap<T> {
 		return this._items.length
 	}
 }
+
+export const enum EnumRunesCode {
+	HIGH_SURROGATE_START = 0xd800,
+	HIGH_SURROGATE_END = 0xdbff,
+
+	LOW_SURROGATE_START = 0xdc00,
+
+	REGIONAL_INDICATOR_START = 0x1f1e6,
+	REGIONAL_INDICATOR_END = 0x1f1ff,
+
+	FITZPATRICK_MODIFIER_START = 0x1f3fb,
+	FITZPATRICK_MODIFIER_END = 0x1f3ff,
+
+	VARIATION_MODIFIER_START = 0xfe00,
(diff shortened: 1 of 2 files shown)


luluxia:
If you set a color for the text, it causes the emoji to be covered by the color, which feels illogical. I made a few more changes so that the color option wouldn't work for emoji characters.

slmjkdbtl:
Thank you this is a very nice feature! I'll take a closer look tomorrow.

src/utils.ts
```diff
@@ -297,3 +297,203 @@ export class BinaryHeap<T> {
...
+	0x0BBF, // ( ி ) TAMIL VOWEL SIGN I
+	0x0BCD, // ( ◌்) TAMIL SIGN VIRAMA
+	0x0E31, // ( ◌ั ) THAI CHARACTER MAI HAN-AKAT
+	0x0E33, // ( ำ ) THAI CHARACTER SARA AM
+	0x0E40, // ( เ ) THAI CHARACTER SARA E
+	0x0E49, // ( เ ) THAI CHARACTER MAI THO
+	0x1100, // ( ᄀ ) HANGUL CHOSEONG KIYEOK
+	0x1161, // ( ᅡ ) HANGUL JUNGSEONG A
+	0x11A8, // ( ᆨ ) HANGUL JONGSEONG KIYEOK
+]);
```
slmjkdbtl, line 340: I'm not knowledge in this field. Is there a reference to all these constants in `EnumRunesCode` and `GRAPHEMES`?

slmjkdbtl:
Is there an existing library you're referencing from?

luluxia:
> Is there an existing library you're referencing from?

yes. the code is from here https://github.com/bluelovers/runes

slmjkdbtl:
It seems legit, gonna merge now.

PR #797, opened 2023-12-03
Update health() in types.ts

luizbills:
The second argument is missing
https://github.com/replit/kaboom/blob/master/src/kaboom.ts#L4445


slmjkdbtl:
thx!

PR #800, opened 2023-12-18
Fixing assets paths in the postEffect example

juneaucross:
Right now demo doesn't work properly :(

image


slmjkdbtl:
The root of the problem is a bug on the website, it's fixed [here](https://github.com/replit/kaboomsite/commit/f708aa7ddd6ecf03794a8b4aba3d4196eb39ae6c). Thanks for the PRs to remind me! But I'm closing them now due to the issue being resolved.

PR #814, opened 2024-01-02
Added support for arrays in uniforms.

tristanperrow:
Arrays for shader uniforms. Soon to support struct like objects to be sent as a uniform. For example, Light a = { vec2 position, vec3 color } will be sent as a uniform such that if your shader has the struct Light, it will be able to accept that uniform to link specific variables together.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### v3000.2.1
+- added support for arrays in uniforms
+
 ## v3000.2
 - added `loadMusic()` to load streaming audio (doesn't block in loading screen)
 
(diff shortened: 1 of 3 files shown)


lajbel:
It looks good for me, @slmjkdbtl check it!

lajbel (approved):

slmjkdbtl:
Thanks, this is good! Might also want to add to the `UniformValue` type [here](https://github.com/replit/kaboom/blob/f27b2220d3977820a32e541e502127bb6dd4adc2/src/types.ts#L4817)

tristanperrow:
Should be good? I don't think I'm missing something, but lmk!

slmjkdbtl:
Thank you, this is great!

PR #815, opened 2024-01-04
Add drawCurve and drawBezier in order to draw curved lines

mflerackers:


src/types.ts
```diff
@@ -3855,6 +3914,8 @@ export declare class Vec2 {
 	toString(): string
 }
 
+export evaluateBezier(pt1: Vec2, pt2: Vec2, pt3: Vec2, pt4: Vec2, t: number): Vec2
```
slmjkdbtl, line 3917: Currently all functions are exported in KaboomCtx [here](https://github.com/replit/kaboom/blob/f27b2220d3977820a32e541e502127bb6dd4adc2/src/kaboom.ts#L6421) even the pure functions

src/math.ts
```diff
@@ -1080,6 +1080,15 @@ export class Polygon {
 	}
 }
 
+export function evaluateBezier(pt1: Vec2, pt2: Vec2, pt3: Vec2, pt4: Vec2, t: number) {
```
slmjkdbtl, line 1083: Thinking about the best name here, what do you think about `generateBezier()`, or `genBezier()`? I'm kinda towards the later

src/kaboom.ts
```diff
@@ -2028,6 +2030,24 @@ export default (gopt: KaboomOpt = {}): KaboomCtx => {
...
+	    }
+	    drawLines({
+	        pts: p,
+	        width: opt.width || 1,
+	        pos: opt.pos,
+	        color: opt.color,
+	    });
+	}
+
+	function drawBezier(opt: DrawBezierOpt) {
```
slmjkdbtl, line 2047: What do you think about `drawBezierCurve()`?

src/math.ts
mflerackers: It literally evaluates the cubic Bezier function at the point t in t-space. It doesn't generate anything other than the point at the given t. I have other functions which evaluate the first derivative (velocity on the Bezier at t, to for example set angle so that you turn along the path), the second derivative (acceleration on the Bezier at t, to animate a thruster for example) or an adapter which evaluates the Bezier in uniform s space, to keep velocity constant, or get fixed offsets along the Bezier or other spline.
genBezier, or better genSpline would be a function which does what goes on partly in drawSpline, evaluating the function in n points to get a list of points representing the spline.

slmjkdbtl: Ok that makes sense 👍

slmjkdbtl:
Thank you! Been wanting to add this.

PR #819, opened 2024-01-28
Add generic shape matching methods + example to test overlap tests

mflerackers:
Shapes now have the following methods

collides(shape: any): boolean - Test overlap with any other shape
contains(point: Vec2): boolean - Test if a point is contained by the shape

### DIFF
--- a/src/math.ts
+++ b/src/math.ts
@@ -1,624 +1,624 @@
 import {
-	Point,
-	RNGValue,
-	LerpValue,
-	Vec2Args,
+    Point,
+    RNGValue,
+    LerpValue,
+    Vec2Args,
 } from "./types"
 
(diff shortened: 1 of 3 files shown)


slmjkdbtl:
Thank you this is useful! Can you fix the diff noise in math.ts it's making it hard to view actual diff

src/math.ts
```diff
@@ -938,6 +1086,12 @@ export class Line {
 	clone(): Line {
 		return new Line(this.p1, this.p2)
 	}
+	collides(shape: any): boolean {
```
slmjkdbtl, line 1089: Maybe give `shape` a union type constraint?

slmjkdbtl:
Thank you!

PR #827, opened 2024-02-26
fix unresolveable collisions in beforePhysicsResolve

neverUsedGithub:
This PR fixes an issue where inside "beforePhysicsResolve" collisions could only be cancelled in the "source" GameObject's listener not inside the "target" GameObject's listener, because the reversed collision's `resolved` property wasn't checked.


slmjkdbtl:
Thank you!

PR #829, opened 2024-03-10
Fix text styles breaking on newlines/wraps

neverUsedGithub:
This PR fixes a bug where when rendering styled text containing newlines or wrapping the index of characters wouldn't line up with their actual index into `charStyleMap`.


slmjkdbtl:
Thank you, nice catch!

PR #833, opened 2024-03-20
Feature/large-texture

csakip:
- When a larger texture is being loaded than the TexPacker was initialized with, it creates a canvas sized to it, stores the texture, then restores the last texture that was being populated.
- Added largeTexture example

### DIFF
--- a/examples/largeTexture.js
+++ b/examples/largeTexture.js
@@ -7,7 +7,7 @@ let cameraScale = 1
 loadSprite("large-image", "https://picsum.photos/2500/2500")
 
 add([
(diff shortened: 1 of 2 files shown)


csakip:
Please note that `npm run lint` gave every line end character as error for all the files, so I couldn't run linting.

slmjkdbtl:
I actually had an implementation for large textures locally but haven't pushed yet, I just [pushed my implementation](https://github.com/replit/kaboom/commit/63fc318705038a21f1cfdda0adf79ab4afdb07fa) but added your example here, thank you!

PR #836, opened 2024-03-26
Added some sprites

SG-SWE073:
spritespatch1
### 3 new sprites:
- A cat
- Snow edge tile
- Snow middle tile
Each of them are 64x64 and have their own commits in case you don't want to use any in particular.


lajbel:
They look good for your own repo! But this is not the style of Kaboom.js sprite right now. We follow a pixel art style with 4px black outline, there's the colour palette: https://lospec.com/palette-list/mulfok32

PR #838, opened 2024-03-26
docs(types.ts): add readd() examples

SG-SWE073:
Saw there was some confusion on how to use readd(), so I added some examples in jsDoc.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -103,6 +103,48 @@ export interface KaboomCtx {
 	make<T>(comps?: CompList<T>): GameObj<T>,
 	/**
 	 * Remove and re-add the game obj, without triggering add / destroy events.
+	 * @example
+	 * ```js
+	 * // Common way to use this is to have one sprite overlap another sprite, and use readd() to have the bottom sprite on top of the other.
+	 * 
+	 * // Create two sprites.
+	 * const greenBean = add([
+	 * 		sprite("bean"),
+	 * 		pos(200,140),
+	 * 		area({ cursor: "pointer" }),
+	 * 		anchor("center"),
+	 * 		scale(2.5),
+	 * 		color(255, 255, 255),  
+	 * ])
+	 * 
+	 * // This bean will overlap the green bean.
+	 * const purpleBean = add([
+	 * 		sprite("bean"),
+	 * 		pos(290,140),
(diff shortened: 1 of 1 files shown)


lajbel:
I feel that the example is too long, could we reduce the components in the example objects? with sprite and pos should be enough

SG-SWE073:
So I updated and shortened some of the unneeded components. This brought up an interesting question to me " Why does it feel too long?", and I think that is because it is near the other examples that are short and don't need much explanation ( like move(), onClick(), fixed() etc.). 
There are some long examples already like state() and body(), that might need some explaining, and imo, the docs should be able to provide that to the users without worrying about the length of the doc or its neighboring examples. 

I think the solution to this would be to only show the current example, or clicked on component from the sidebar, on its own page, so A) the explanation can be as long as needed. B) The user isn't seeing everything all at once ( doom scroll ).

Here are some examples of what I mean:
Three.js creating text has it's own page: https://threejs.org/docs/index.html#manual/en/introduction/Creating-text
Unity lighting is subdivided with their own pages: https://docs.unity3d.com/Manual/choose-a-lighting-setup.html

I know there are other issues that should be taken care of first, but maybe this can be considered for v4000-v5000.

SG-SWE073:
Does the code need to be adjusted any further, or is it good to go?

src/types.ts
```diff
@@ -103,6 +103,41 @@ export interface KaboomCtx {
...
+	 * @example
+	 * ```js
+	 * // Common way to use this is to have one sprite overlap another sprite, and use readd() to have the bottom sprite on top of the other.
+	 * 
+	 * // Create two sprites.
+	 * const greenBean = add([
+	 * 		sprite("bean"),
+	 * 		pos(200,140),
+	 * 		color(255, 255, 255),
+         *		area(),
```
lajbel, line 115: There's an extra indentation?

lajbel (changes requested):

lajbel:
Is looking good for me, only that little thing

SG-SWE073:
Indents should be removed now with recent commit

lajbel:
Looking good!

lajbel (approved):

lajbel:
Thank you for your contribution to documentation! 😇

PR #841, opened 2024-03-31
Feat: Object right-click

Meldiron:
Adds ability to configure type of click for object `onClick` callback.


slmjkdbtl:
Looks good! Remember to also update corresponding type def in `types.ts`

amyspark-ng:
very good

PR #842, opened 2024-04-04
Improved docs regarding custom events usage

JSLegendDev:
In the Kaboom.js docs, it's not clear how to use the `on()` function used to define custom events.
I tried to improve the docs to make usage more clear.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -672,9 +672,12 @@ export interface KaboomCtx {
 	 * // a custom event can be defined manually
 	 * // by passing a name and a callback function
 	 * on("talk", (message, posX, posY) => {
-	 *     add([text(message), pos(posX, posY - 100)])
+	 *     add([
+	 * 		text(message), 
+	 * 		pos(posX, posY - 100)
+	 * 	   ])
 	 * })
-	 * onKeyPress("Space", () => {
+	 * onKeyPress("space", () => {
 	 *    	// the trigger method on game objs can be used to trigger a custom event
 	 * 		npc.trigger("talk", "Hello World!", npc.pos.x, npc.pos.y)
 	 * })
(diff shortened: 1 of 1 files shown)


lajbel:
I would also mention that there's a list of avaible events, like ground for Body, we should expose it in types too!

lajbel (changes requested):
You should add a mention here about that there's no only custom events, there's also default events, ground is one, and there's others. So maybe I would change it and also mention what are the defaults events.

It's a fact that we don't expose the events very well, maybe another PR for the repo

JSLegendDev:
I opted to just add to the existing doc that you can have custom events with an example instead of removing what was there before.

src/types.ts
```diff
@@ -668,6 +668,17 @@ export interface KaboomCtx {
 	 *     destroy(bomb)
 	 *     addKaboom(bomb.pos)
 	 * })
+	 * 
+	 * // a custom event can be defined manually
+	 * // by passing a name and a callback function
+	 * on("talk", (message, posX, posY) => {
```
lajbel, line 674: indentation on every component here

src/types.ts
```diff
@@ -668,6 +668,17 @@ export interface KaboomCtx {
...
+	 *     add([text(message), pos(posX, posY - 100)])
+	 * })
+	 * onKeyPress("Space", () => {
```
lajbel, line 677: Space in lowercase

lajbel (changes requested):
Only a few things, but looking good

JSLegendDev:
@lajbel Addressed feedback and fixed spacing issues.

lajbel (approved):
looking good!

lajbel:
Thank you for your contribution to documentation! 🔥

PR #849, opened 2024-04-18
remove docs of removed components, simplify readd() example

lajbel:
- Remove mention to `children()` and `solid()` components
- Simplify the `readd()` example

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -92,12 +92,8 @@ export interface KaboomCtx {
 	 * const label = make([
 	 *     text("oh hi"),
 	 * ])
-
-	 * add([
-	 *     rect(label.width, label.height),
-	 *     color(0, 0, 255),
-	 *     children(label),
-	 * ])
+	 * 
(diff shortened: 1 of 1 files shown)


src/types.ts
```diff
@@ -92,51 +92,30 @@ export interface KaboomCtx {
...
 	 *     text("oh hi"),
 	 * ])
-
-	 * add([
-	 *     rect(label.width, label.height),
-	 *     color(0, 0, 255),
-	 *     children(label),
-	 * ])
+	 * 
+	 * add(label);
```
slmjkdbtl, line 96: this part is to have an example text with background which is made possible

lajbel: Any update @slmjkdbtl

slmjkdbtl: yeah something like

```js
const label = make([
    text("hi")
])

const background = add([
    rect(label.width, label.height)
])

background.add(label)
```

which might be a more practical example

PR #855, opened 2024-05-05
Update Kaboom unpkg version in setup doc

jpg-gamepad:
The tutorial is linking to an older version of Kaboom


lajbel (approved):
Thank you!

PR #844, opened 2024-04-06
fadeIn() now uses opacity to figure out the final opacity to fade into

niceEli:
Changed fadeIn to use opacity so it can fade into 0.5 for example
Changed Contributing.md because there was "" around linebreak style and that isnt correct
Added fadeIn example so you can learn fadeIn
Changed JSDoc to have a fadeIn example

PR #858, opened 2024-05-08
changelog added `fadeIn()` now uses `opacity()` to determinate the final opacity

lajbel:


lajbel:
This should be a breaking change? 

https://github.com/replit/kaboom/pull/844#issuecomment-2100072427
> Also, would this be considered a breaking change since all users who specify `opacity()`, then `fadeIn()` now will see the opacity actually being applied.

lajbel:
@slmjkdbtl any updates?

